PR #91, opened 2026-03-16
Proposal: Daml upgrade migration planner

pmensik:

**Proposal file:** `/proposals/daml-upgrade-migration-planner.md`

---

## Summary
This proposal introduces the **Daml Upgrade Migration Planner**, a CLI-first common-good tool designed to automate and derisk smart contract upgrades on Canton. It connects to a live participant, evaluates existing package and vetting state against newly proposed DARs, and computes a deterministic, reviewable migration plan mapping out exact upload orders, multi-synchronizer vetting steps, and topology serials for safe rollback.

---

## Checklist

---

## Notes for Reviewers
- **Focus on tooling over APIs:** The tool wraps and orchestrates existing Canton primitives (`dpm upgrade-check`, `ValidateDarFile`, and Admin API `dry_run`) into a single planning workflow rather than re-inventing Daml-LF validation.
- **Bounded Scope:** The tool strictly focuses on single-participant migration *planning* through JSON/Markdown output. Distributed multi-participant coordination and mutative release orchestration are explicitly framed as follow-up grant opportunities. 
- **Operational Realism:** The design natively recognizes Canton's rollback asymmetry, planning for the fact that package uploads are irreversible, but vetting topology changes can be safely rolled back using CAS topology serials.


hythloda:
## Review Feedback — Daml Upgrade Migration Planner

Thank you for the submission and for contributing a proposal focused on improving upgrade safety and operational workflows for Daml applications on Canton.

This proposal is well-scoped, technically grounded, and addresses a clear operational pain point around upgrade sequencing, vetting, and rollback planning. 
That said, a few areas would benefit from clarification:

---

### 1. Evidence of ecosystem demand

The proposal clearly outlines real failure modes, but it would help to clarify:
- whether multiple teams are currently experiencing these upgrade challenges in practice  
- any examples of teams building internal tooling or manual workflows today  

---

### 2. Adoption expectations

It would be helpful to clarify:
- who the primary users are (e.g., app teams, validator operators, SREs)  
- whether there are early adopters or teams planning to integrate this into CI/CD workflows  
- what success looks like in terms of usage across the ecosystem  


Overall, this is a strong and well-defined proposal that targets a real operational gap. Clarifying demand and adoption pathways would strengthen confidence in ecosystem-wide impact.

pmensik:
Hello @hythloda, I am sorry for the late reply. 
Thank you for the feedback — both points are fair and makes sense.

## 1.  Evidence of demand/ecosystem validation

We did try to validate this directly by reaching out to several teams building on Canton, but unfortunately we did not get meaningful responses. There also does not currently seem to be an official developer forum or a dedicated Discord channel suitable for this kind of tooling poll, so broader validation was difficult to do in a structured way.
Because of that, this proposal is primarily based on our own hands-on experience experimenting with Canton and on the pain points we encountered around upgrade planning, package sequencing, vetting coordination, and rollback preparation. We assume that teams facing similar upgrade workflows are likely handling them today with internal runbooks or ad hoc scripts rather than with a dedicated tool.

## 2. Intended users / likely adoption
The primary users we have in mind are Daml application teams and platform or release engineers responsible for planning and executing package upgrades safely. This is not intended as a broad onboarding tool, but rather as a focused operational tool for teams dealing with non-trivial upgrade workflows.

The main users we have in mind are:
 - Daml application teams preparing package upgrades,
 - platform / SRE teams responsible for release workflows,
 - operators who need a deterministic and reviewable migration plan before execution.

It is less targeted at general validator operations and more at teams directly managing Daml package rollout safety.

Denend:
@pmensik 
Plan Engine assigns preconditions/warnings to Vet/Unvet steps based on per-package, per-synchronizer vetting state.

Does it account for the SCU compatibility exemption - a package doesn't need direct vetting if a compatible upgrade is vetted - and the unvet precondition — all v1 contracts must be fully upgraded before v1 can be unvetted?

Ref: https://docs.canton.network/appdev/modules/m6-package-selection.md - "A contract created with V1 of a template can be used/upgraded with/to V2, even though V1 was unvetted provided that V2 is vetted." / "A participant node must fully upgrade all its v1 contracts before unvetting v1."

pmensik:
@Denend Hello Daniel, thank you for your feedback, this is a very good point. 

It's true that the current proposal does not describe this accurately enough. The planner should account for package selection and upgrade compatibility, not just raw package/vetting state. In particular, V2 being vetted can change whether V1 still needs to be directly vetted, but unvetting V1 should only be considered safe once the remaining V1 contracts have been archived or migrated. This is what I had in my mind while writing the proposal, however, it feels like I'll need to update it with better wording

askardex:
At Askardex, we really need something like this. 

Our main struggle right now is dealing with SV vetting lag. We often upgrade our Daml contracts to match the latest Splice versions, only to find out too late that the SV hasn't vetted them yet.

Right now, checking this manually is a hassle, and the compatibility dashboard isn't always perfectly accurate for us. So we end up upgrading, hitting a wall, and then having to downgrade all over again, which is super annoying.

If this planner can just check our target packages against the actual network state and give us a simple "safe to upgrade" or "do not upgrade yet," it would save our team a lot of time.
Fully support this proposal!

leonidr-c7:
Interesting and potentially very useful tool. While this is not a big problem for C7 at the moment, I do foresee a future with complicated migrations. Some thoughts:

1. I would personally prefer that this be directly integrated into dpm as opposed to a standalone tool. I will forget the name as upgrading is not a frequent occurrence.
2. Support for different nets. For DevNet and TestNet, we create dars that we never upload to MainNet. Sometimes we manually just clear the state to allow the dars to be updated and vetted on that net. Having a different strategy per net would be useful, even better would be an integrated capability where testing on TestNet given plan `a`, gives you confidence that plan `b` for MainNet will work.
3.  There are workflows that it would be nice to support that are really outside the realm of traditional SCU workflow. Some users might add automation to orchestrate updates to contracts naturally as they roll over so that one does not have to generate excess traffic. For example, I could get a counterparty to agree to another contract to delegate the right to upgrade to a non-SCU compatible workflow. With this in mind, it would be nice if the 'state' of the upgrade was something that the tool could track (integrate with PQS to track progress, show old and new versions). This was the original suggested upgrade path I used to give to clients before SCU. There are a couple of parts of this request:

   - More refined, specific upgrade path.
   - Tracking this state against existing contracts.
   - Delegated automation to include an upgrade when one does have to take an action to cut transaction costs.


Thank you @pmensik !

pmensik:
> At Askardex, we really need something like this.
> 
> Our main struggle right now is dealing with SV vetting lag. We often upgrade our Daml contracts to match the latest Splice versions, only to find out too late that the SV hasn't vetted them yet.
> 
> Right now, checking this manually is a hassle, and the compatibility dashboard isn't always perfectly accurate for us. So we end up upgrading, hitting a wall, and then having to downgrade all over again, which is super annoying.
> 
> If this planner can just check our target packages against the actual network state and give us a simple "safe to upgrade" or "do not upgrade yet," it would save our team a lot of time. Fully support this proposal!

Hello Irpan, thanks a lot for your feedback, looks like the migration planner would be really helpful for your case!

pmensik:
> Interesting and potentially very useful tool. While this is not a big problem for C7 at the moment, I do foresee a future with complicated migrations. Some thoughts:
> 
> 1. I would personally prefer that this be directly integrated into dpm as opposed to a standalone tool. I will forget the name as upgrading is not a frequent occurrence.
> 2. Support for different nets. For DevNet and TestNet, we create dars that we never upload to MainNet. Sometimes we manually just clear the state to allow the dars to be updated and vetted on that net. Having a different strategy per net would be useful, even better would be an integrated capability where testing on TestNet given plan `a`, gives you confidence that plan `b` for MainNet will work.
> 3. There are workflows that it would be nice to support that are really outside the realm of traditional SCU workflow. Some users might add automation to orchestrate updates to contracts naturally as they roll over so that one does not have to generate excess traffic. For example, I could get a counterparty to agree to another contract to delegate the right to upgrade to a non-SCU compatible workflow. With this in mind, it would be nice if the 'state' of the upgrade was something that the tool could track (integrate with PQS to track progress, show old and new versions). This was the original suggested upgrade path I used to give to clients before SCU. There are a couple of parts of this request:
> 
> * More refined, specific upgrade path.
> * Tracking this state against existing contracts.
> * Delegated automation to include an upgrade when one does have to take an action to cut transaction costs.
> 
> Thank you @pmensik !

Hello Leonid, thank you for your feedback, really appreciated! To answer each of your concern separately
1) I agree that this would be the desired output, however, in the first iterations it's easier to reason about the tool when it is standalone. Also it would require coordination with the Digital Assets Holdings on the integration which would slow down the initial development. So yeah, this is definitely where I would like to see it going in the future but I expect it to remain standalone before it can prove itself in production setups. 
2) Definitely makes sense and should be pretty easy to add - I'll update the proposal along these lines
3) Interesting suggestions, I have to admit I didn't think about all of them when writing the proposal. 
   a) Yes, introducing various migration strategies based on real upgrade issues sounds useful
   b) Gathering contract information and their state from PQS could indeed go even in the first milestones while adding more interactivity later on
   c) Yeah, automation or deferred execution would be interesting but that's most likely the most complex feature to deliver

Denend:
@pmensik the discussion has landed several agreed changes (the SCU vetting-compatibility handling, RecordTime/config points, multi-net strategy, and the dpm-integration direction), but the proposal file hasn't been updated since the original submission. Could you push a revision folding those in?

sunstrike228:
Read the full proposal properly. This is terraform plan for Daml upgrades, and that is the shape this problem needed. The part that stands out most is planning around the rollback asymmetry: package uploads are irreversible, vetting changes are not, and that distinction is exactly where live upgrades go wrong. Producing a reviewable plan with exact upload order, per-synchronizer vetting steps and topology serials before anything touches the participant is how operators already think, just without tooling to back it up. Also sensible that it orchestrates `dpm upgrade-check`, `ValidateDarFile` and the Admin API `dry_run` rather than reimplementing validation, which means the plan stays honest as Canton itself evolves. Keeping v1 to single-participant planning with JSON and Markdown output, and leaving mutative orchestration to a follow-up, reads like discipline rather than a gap. Every app provider shipping Daml upgrades ends up hand-writing a worse version of this in runbooks and shell scripts, so a common-good version is easy to support.

hgrano:
I think this would be a genuinely useful tool. For some upgrades, this tool (at least in its initial form) won't work, because they require executing custom logic on the ledger. For example, propose-accept patterns to archive old contracts and create new versions. Some way to track the progress of these more complex upgrade processes would be useful but beyond the scope of this proposal.

pmensik:
I’ve updated the proposal based on the review feedback and subsequent technical discussion.

The main changes are:

* SCU/package-selection-aware vetting and unvetting planning
* active-contract and compatible-replacement analysis
* explicit state provenance
* scoped package-support reporting for the local participant, synchronizers, and configured external participants
* named DevNet/TestNet/MainNet profiles and plan comparison
* clearer rehearsal, rollback, visibility, and non-goal boundaries
* future `dpm` integration direction

I also updated the milestones and acceptance criteria to reflect the additional work, extending delivery to 17 weeks and increasing the funding request by roughly 10% to cover for the additional work.

Some features (like reading contract state from PQS or deferred execution) were intentionally omitted in the first version - I wanted to keep our proposal lean and showcase our ability to quickly deliver a working product. But obviously there are already many features that would make a great candidate for v2.

Brindrajsinh-Chauhan:
Thanks for putting this together — after going through the proposal, this looks like a useful tool that teams could adopt as a validation gate in their pipelines. The core idea of enforcing checks and compatibility before changes land is solid, and it clearly overlaps with problems that infra-as-code tooling (Terraform in particular) is already trying to solve.
One thing we think would be useful to add to this proposal is a section comparing it against existing workflow/automation options that were considered, so reviewers can see where it provides distinct value rather than overlapping capability. A few potential candidates worth covering:

**Custom Terraform Provider**
A well-maintained custom provider could implement much of this validation logic and plug directly into teams' existing resource-management pipelines — no new tool to learn, no separate system to integrate, just an import away for teams already on Terraform. That said, this proposal's ability to hook into existing dpm tooling and validator checks gives it a real edge here: a Terraform provider would likely need to reimplement or wrap that logic rather than reuse it natively. The tool's ability to use existing protobufs to prevent drift is also a plus — a custom provider would require continuous effort to keep its protobuf definitions in sync with upstream, whereas this tool seems to avoid that maintenance burden entirely based on this proposal.

**GitHub Actions / CI workflow tools**
GitHub Actions (or similar CI-native tooling) is the other natural comparison, since it's already where most teams gate merges today. It's flexible and requires no new infrastructure, but validator-specific logic tends to end up scattered across YAML and shell scripts per repo, with little reuse or central ownership. There's also a safety consideration: Actions workflows typically need direct write access to connect to and act on production systems, which raises the blast radius if a workflow is misconfigured or a script has a bug — the risk of an action causing an unintended production change is real. This tool enforcing read-only access by default, with explicit gates required before anything can touch production, is a meaningful advantage — it reduces the chance of accidental production changes compared to a CI pipeline with standing write access.

Adding a section like this — and including more comparisons along these lines — would strengthen the proposal by making the "why this over what we already have" case explicit, rather than leaving reviewers to infer it

pmensik:
I added another update of the proposal based on recent reviews and ecosystem feedback
- Added new M1 to represent a working vertical-slice to validate the architecture before full implementation
- Added lightweight design-partner validation after M1, with public, anonymized, and confidential participation options
- Added a final adoption milestone covering independent production or production-equivalent evaluations
- Strengthened the `dpm` integration path and introduced early maintainer consultation
- Added clear maintenance, compatibility, and stewardship commitments
- Kept advanced functionality such as PQS tracking and execution outside v1 
- Rebalanced milestone funding while keeping the total request unchanged at 585,000 CC

waynecollier-da:
> At Askardex, we really need something like this.
> 
> Our main struggle right now is dealing with SV vetting lag. We often upgrade our Daml contracts to match the latest Splice versions, only to find out too late that the SV hasn't vetted them yet.
> 
> Right now, checking this manually is a hassle, and the compatibility dashboard isn't always perfectly accurate for us. So we end up upgrading, hitting a wall, and then having to downgrade all over again, which is super annoying.
> 
> If this planner can just check our target packages against the actual network state and give us a simple "safe to upgrade" or "do not upgrade yet," it would save our team a lot of time. Fully support this proposal!

Are you saying you upgrade to new Daml models before the SVs announce that they have vetted a new Splice version, and before they have adopted them? This information is published by the SVs, so it's hard to see how it would be a manual step.   

Could you spell out the sequence of steps here more precisely?

pmensik:
> I think this would be a genuinely useful tool. For some upgrades, this tool (at least in its initial form) won't work, because they require executing custom logic on the ledger. For example, propose-accept patterns to archive old contracts and create new versions. Some way to track the progress of these more complex upgrade processes would be useful but beyond the scope of this proposal.

@hgrano Hello Huw, thanks a lot for your feedback. It's a deliberate intention to keep this proposal rather small while providing as much utility as it can. There are many directions that we would like to explore in the future (PQS state inspection,  user-defined workflows or automated workflow execution). However, adoption must come first and therefore we want to use funding to build a MVP that would already be useful, gather feedback and then explore more advanced features.

pmensik:
> Thanks for putting this together — after going through the proposal, this looks like a useful tool that teams could adopt as a validation gate in their pipelines. The core idea of enforcing checks and compatibility before changes land is solid, and it clearly overlaps with problems that infra-as-code tooling (Terraform in particular) is already trying to solve. One thing we think would be useful to add to this proposal is a section comparing it against existing workflow/automation options that were considered, so reviewers can see where it provides distinct value rather than overlapping capability. A few potential candidates worth covering:
> 
> **Custom Terraform Provider** A well-maintained custom provider could implement much of this validation logic and plug directly into teams' existing resource-management pipelines — no new tool to learn, no separate system to integrate, just an import away for teams already on Terraform. That said, this proposal's ability to hook into existing dpm tooling and validator checks gives it a real edge here: a Terraform provider would likely need to reimplement or wrap that logic rather than reuse it natively. The tool's ability to use existing protobufs to prevent drift is also a plus — a custom provider would require continuous effort to keep its protobuf definitions in sync with upstream, whereas this tool seems to avoid that maintenance burden entirely based on this proposal.
> 
> **GitHub Actions / CI workflow tools** GitHub Actions (or similar CI-native tooling) is the other natural comparison, since it's already where most teams gate merges today. It's flexible and requires no new infrastructure, but validator-specific logic tends to end up scattered across YAML and shell scripts per repo, with little reuse or central ownership. There's also a safety consideration: Actions workflows typically need direct write access to connect to and act on production systems, which raises the blast radius if a workflow is misconfigured or a script has a bug — the risk of an action causing an unintended production change is real. This tool enforcing read-only access by default, with explicit gates required before anything can touch production, is a meaningful advantage — it reduces the chance of accidental production changes compared to a CI pipeline with standing write access.
> 
> Adding a section like this — and including more comparisons along these lines — would strengthen the proposal by making the "why this over what we already have" case explicit, rather than leaving reviewers to infer it

Thanks, this is a useful suggestion. We agree the proposal should explain the alternatives more explicitly.

Our current view is that `dpm` should be the primary integration point, while GitHub Actions and Terraform are execution or orchestration layers rather than separate implementations. If the planner exposes a stable, non-interactive CLI with JSON output and deterministic exit codes, CI systems can run it directly without a dedicated integration. Terraform support would only make sense later if there is concrete demand.

pmensik:
@Denend I added the following changes based on your feedback 
- Added the complete exact `UpdateVettedPackagesForceFlag` enum and modeled unvetted dependency consistency separately from active-contract/SCU safety.
- Narrowed the documented scope of `ValidateDarFile` to participant-side DAR and upgrade-compatibility validation.
- Added explicitly configured party-scoped readiness using `GetPreferredPackages`, with host-level topology diagnostics where visible. This remains scoped to the participant’s observed topology, not global readiness.
- Clarified that V1 remains standalone, while third-party opt-in `dpm` packaging is possible; official bundling still requires maintainer agreement. TestNet evidence is never promoted into a MainNet safety verdict.
- Rebalanced funding from M4 into M1/M2 to make cover increased amount of work in M1/M2
- Strengthened M1 with focused JVM/GraalVM parity and a production-path DAR/DALF + SCU planning proof, without moving the full M2–M4 scope into M1.

Denend:
@pmensik really solid work, and the SCU/vetting handling is spot on. The M1 slice and the design-partner gate were good additions. A few more issues:

1. Funding split. M6 is 45,000 of 585,000, so 7.7% sits behind adoption. Across the thirteen approved proposals that price an adoption milestone the range is 20% to 100%, median 43.5%, and the lowest in daml-tooling is 30%. Could you shift enough into M6 to land around 30%, and split M3 so no milestone carries a third of the budget?

2. @waynecollier-da's question from 29 July on the SV vetting-lag sequence is still open. That is the concrete demand datapoint in the thread, worth closing out.


Minor: the rollback section treats package uploads as irreversible, but RemoveDar and RemovePackage exist on the Admin API with documented preconditions.

refs
CIP-0100, Governance Process 2.2: https://github.com/canton-foundation/cips/blob/main/cip-0100/cip-0100.md

pmensik:
> @pmensik really solid work, and the SCU/vetting handling is spot on. The M1 slice and the design-partner gate were good additions. A few more issues:
> 
> 1. Funding split. M6 is 45,000 of 585,000, so 7.7% sits behind adoption. Across the thirteen approved proposals that price an adoption milestone the range is 20% to 100%, median 43.5%, and the lowest in daml-tooling is 30%. Could you shift enough into M6 to land around 30%, and split M3 so no milestone carries a third of the budget?
> 2. @waynecollier-da's question from 29 July on the SV vetting-lag sequence is still open. That is the concrete demand datapoint in the thread, worth closing out.
> 
> Minor: the rollback section treats package uploads as irreversible, but RemoveDar and RemovePackage exist on the Admin API with documented preconditions.
> 
> refs CIP-0100, Governance Process 2.2: https://github.com/canton-foundation/cips/blob/main/cip-0100/cip-0100.md

1. For the funding split - we originally had another adoption gate after M3 where further development was dependent on validation of the design by the design partners. But I agree that M3 was too big while more of the funding needed to move behind real adoption. Therefore, we moved 175,000 CC (29.9%) behind adoption and split the plan-engine work into two milestones. Adoption payments are now proportional to verified evaluations and operational use. No milestone exceeds 30%, and the total remains 585,000 CC.

2. I pinged @askardex if they can provide an answer to that since this about their workflow mostly

askardex:
> > At Askardex, we really need something like this.
> > Our main struggle right now is dealing with SV vetting lag. We often upgrade our Daml contracts to match the latest Splice versions, only to find out too late that the SV hasn't vetted them yet.
> > Right now, checking this manually is a hassle, and the compatibility dashboard isn't always perfectly accurate for us. So we end up upgrading, hitting a wall, and then having to downgrade all over again, which is super annoying.
> > If this planner can just check our target packages against the actual network state and give us a simple "safe to upgrade" or "do not upgrade yet," it would save our team a lot of time. Fully support this proposal!
> 
> Are you saying you upgrade to new Daml models before the SVs announce that they have vetted a new Splice version, and before they have adopted them? This information is published by the SVs, so it's hard to see how it would be a manual step.
> 
> Could you spell out the sequence of steps here more precisely?



Thanks for the clarification. To be precise, our problem is not that we intentionally upgrade before an SV announces support.

The issue is the operational verification step after a new Splice/Daml version becomes available. Our usual sequence is:

1. A new Splice/Daml release or DAR is published.
2. We upgrade or prepare an updated DAR for our application.
3. We upload the new DAR and its dependencies to our validator participant.
4. The application fails because one or more packages are not yet vetted on the synchronizer, or because the required dependency version is not available there.
5. We then have to determine whether the problem is caused by:
   - the DAR not being uploaded,
   - the package not being vetted,
   - the package not being synchronized-vetted on the active synchronizer,
   - a missing or incompatible dependency, or
   - an application contract still referencing the previous package version.
6. If the upgrade cannot proceed, we have to roll back or keep the previous DAR version active.

The difficulty is that checking only the local participant package list is not sufficient. A package may be uploaded but not vetted on the active synchronizer. In some cases, the compatibility dashboard also does not make this distinction clear enough for our deployment workflow.

What would help us is a planner that checks the exact target package IDs and dependency versions against the actual participant and synchronizer state, including synchronized vetting and upgrade compatibility, and then reports something like:

- Safe to upgrade
- Uploaded but not vetted
- Dependency missing
- Vetted but not compatible
- Keep current version

For example, when we recently prepared a Gift DAR upgrade, the correct validation required checking both DAR package presence and the active synchronizer's vetted package state. The package was safe only after those checks passed, while the previous version still had to remain installed because existing contracts referenced it.

pmensik:
Hello @waynecollier-da, do you have other questions about the proposal?

pmensik:
I have updated the proposal to be aligned with Dev Fund 2.0. RFP

PR #99, opened 2026-03-17
Proposal: Curated Lending and Tokenized Vault Standard on Canton

Kelmat-ai:

**Proposal file:**  
/proposals/mystic.md

---

## Summary
Mystic proposes to build a) an open-source vault standard and b) a curated lending market on Canton, which enables curators to create isolated lending vaults which allocate capital according to their risk preferences. This means isolated risk exposure for LPs and increased capital efficiency for borrowers, all under the privacy-preserving stack of Canton.

---

## Checklist

---

## Notes for Reviewers
- Champion: @gabitu7 (Bitsafe)

- We're pretty much ready with the protocol's milestone 1. Audits (and thus, this proposal) are about to become our main blocker from going to market on Canton.


Kelmat-ai:
AMMENDMENT TO PROPOSAL:

PROBLEM:
As we're wrapping up development of our protocol's markets layer, we've begun focusing on the development of curated vaults on Canton. That's when we've realized that there is no standard CIP yet for vaults on Canton, similar to the ERC-4626 standard on the EVM. We've asked around and are seeing multiple players in the ecosystem working on their own version of ERC-4626-equivalent vaults at the same time, meaning the chain will have multiple vault standards in production. We are about to do the same thing, as we need vaults for our protocol. This will be very detrimental for Canton, as protocols will need to conform to many standards and implement their own adapters for each one, making it difficult for projects to integrate and collaborate with each other if they don't have a common vault standard to follow.

SOLUTION:
As such, we hereby would like to ammend this proposal to add that we will, as part of this proposal, present a CIP proposing an open standard for ERC-4626-equivalent vaults on Canton, which can be used by anyone. This will highly benefit the ecosystem and prevent the fragmentation we're currently seeing that will be really negative for DeFi on Canton.

IMPLEMENTATION MECHANICS:
What we're proposing is a uniform Daml interface that accepts a single CIP-0056 token as deposit and issues another CIP-0056 token as vault share. Redemptions are processed by burning vault shares and, in exchange, give users their initial deposit assets. The implementation we have in mind features 16 methods on a VaultFactory interface, modeled after the ERC-4626 standard. They are:
- 4 core methods - Vault_Deposit, Vault_Mint, Vault_Withdraw, Vault_Redeem. These change the state of the vault and add/subtract from existing assets.
- 4 "max" methods - Vault_MaxDeposit, Vault_MaxMint, Vault_MaxWithdraw, Vault_MaxRedeem. These check vault caps on every method being executed.
- 4 preview methods - Vault_PreviewDeposit, Vault_PreviewMint, Vault_PreviewWithdraw, Vault_PreviewRedeem. These allow the user to simulate the outcome of the method they are about execute.
- 2 oracle methods - Vault_ConvertToShares, Vault_ConvertToAssets. These price the exchange of vault shares to assets provided.
- 2 read methods - UnderLyingAssets and TotalAssets. These let others read vault-specific information.
We assume every vault deployer will create an API that enables external parties to not only read information, but exercise the above methods when allowed to. This will enable parties that otherwise don't have visibility over the vault's contracts to interact with it. We will submit a template for this API also.

MOTIVATION:
Ultimately, a thriving DeFi ecosystem lives on modularity and interoperability. Having multiple vault standards will seriously hinder both of these, introduce a lot of confusion and ultimately hurt the growth of the ecosystem. As someone that needs this CIP to exist and would have liked to have one now, we don't want to see this happen. As such, we instead propose we build one instead so that future developers don't face the same blocker.

RATIONALE:
Mystic has to build curated lending vaults as part of doing curated lending on Canton. Meaning we will build part of this ourselves with or without doing the CIP. As such, we're in the perfect position to take this on, as it will represent less added work for us than anyone else in the ecosystem. In addition, we know exactly what a developer building a vault-related product would need to see in a CIP for it to be helpful for them, much like we're in contact with the other parties doing their own vault standards and can thus align on something that is good for everyone. We are thus in the "right place, right time" to be doing this vault standard.

FINAL NOTES:
- Please assume that the vault standard will be delivered and the corresponding CIP submitted as part of our first Milestone. Submitting the CIP can be an added acceptance criteria for the first milestone, should you accept this.
- This is an add-on to the existing proposition, and we propose we do it in tandem with it.

We hope you will consider this and we reiterate the need for this standard for the DeFi ecosystem, as we're already seeing multiple people that would benefit from this and will only see more over time. CC'ing you in case you had already seen the initial proposal @hythloda @tkatrichenko @stas-sbi @waynecollier-da @isegall-da. Thanks!

waynecollier-da:
It seems reasonable to add a CIP proposal for ERC-4626-comparable vaults. I'd encourage you to do that.  It also strengthens your proposal, since it allows you to create some OSS artifacts for the community. And you might want to include an OSS reference implementation of that vault standard as part of your scope. 

For the rest of the scope, I'd encourage you to more clearly call out the audits in your milestones and payment schedule: when will they be performed, what scope will they cover, and how the audit firm(s) will be paid (by you? by the Foundation directly via a rewards coupon?)

Kelmat-ai:
Thanks for the quick response @waynecollier-da , we appreciate the feedback.

We've gone ahead and done as you suggested: edited the proposal so it includes an Open-Source reference implementation as a deliverable, alongside the CIP proposal.

We've also simplified the proposed development project (we have more experience with DAML now and know better what the work involved for each Milestone is). In addition, also as you have suggested, we have introduced audit timelines, scopes and payment details. We've even added the auditors we'd like to work with. Hope all is clear, thanks again for reviewing.

proposals/mystic.md
```diff
@@ -0,0 +1,239 @@
...
+* We will pay the audits ourselves from the funding received from this milestone. Meaning, we are here assuming that the funding is disbursed after the acceptance criteria is confirmed and it has been validated that the code is ready for auditing.
+
+**Deliverables:**
+* 3 MysticMarket Daml smart contracts - base.dar, curator.dar and oracle.dar
+* Market functionality on Devnet- supply, withdraw, borrow, repay, liquidate. IRM, LTV/LLTV and custodian implementations all working flawlessly;
+* Pass end-to-end testing of all edge cases and internal testing;
+* Internal audit and internal review;
+* UI and backend integrations so markets are functional and available on Devnet in the Mystic UI.
+
+## **Milestone 2: Vaults**
```
VinyoDA, line 95: What about ERC-7540?

proposals/mystic.md
```diff
@@ -0,0 +1,239 @@
...
+- The open-source ERC-4626-like vault standard we've proposed building is in direct alignment and will be integrated with the [Decentralization Manager](https://github.com/canton-foundation/canton-dev-fund/pull/298) proposed by Bitsafe in PR 298 to the Dev Fund, as vaults will be able to choose having centralized custody or use the Decentralization Manager for decentralized custody.
+- We're building everything based on CIP-56 for asset standards at both vault and market level and the oracle template at the market level.
+
+### 4. Backward Compatibility
+
+No backward compatibility impact.
+
+## Milestones and Deliverables
+
+## **Milestone 1: Markets**
```
VinyoDA, line 64: Clear indication on what's going to be OSS and what is going to remain proprietary with Mystic.

proposals/mystic.md
```diff
@@ -0,0 +1,239 @@
...
+
+The proposal also aligns with ecosystem priorities to further interoperability and broaden participation, as Mystic advances Canton DeFi modularity and increases DeFi addressable market dramatically by enabling many more issuers to be onboarded as lending collateral and supply assets, as well as enabling anyone to build private and isolated lending experiences on top of Mystic. The proposed introduction of a unified vault standard as a new CIP is also fully in alignment with these objectives.
+
+More specifically, in terms of alignment with existing ecosystem initiatives and CIPs:
+- The open-source ERC-4626-like vault standard we've proposed building is in direct alignment and will be integrated with the [Decentralization Manager](https://github.com/canton-foundation/canton-dev-fund/pull/298) proposed by Bitsafe in PR 298 to the Dev Fund, as vaults will be able to choose having centralized custody or use the Decentralization Manager for decentralized custody.
+- We're building everything based on CIP-56 for asset standards at both vault and market level and the oracle template at the market level.
```
VinyoDA, line 56: Would be great to take token standard v2 into consideration already.

proposals/mystic.md
```diff
@@ -0,0 +1,239 @@
...
+
+**Timeline**: 3 months from Milestone 2 completion.
+
+From a distribution perspective, we plan to have two curators initially: one for CC and one for stablecoins, each with their own distribution advantage in the respective assets. We will work with curators and issuers on the chain to set up markets which can enable CC-denominated and stablecoin-denominated leverage loops. To add that a third vault targeting a  CBTC-denominated loop is already in the works as well.
+
+We will bootstrap supply-side TVL with CC rewards and a CEX integration we have closed which will direct users to supply their assets on Mystic. From there, we will work with issuers and curators to attract borrowers, effectively getting the flywheel to start turning. After that, our focus will be on onboarding more issuers, curators, Canton validators and distribution partners to scale the number of and size of vaults available on Canton.
+
+---
+
+## Funding
```
VinyoDA, line 177: What about maintenance and code ownership?

proposals/mystic.md
Kelmat-ai: We wont use ERC-7540-equivalent logic on our lending market, nor is it part of canonical ERC-4626 on the EVM, so we weren't thinking of adding that to the OSS CIP either. That said, 100% it's crucial for stuff like strategy vaults to be possible, much like issuers will benefit of it.

Given how 112 already supports the 7540 behaviour up to a certain extent, a good way forward here could be adding that to the vault standard as an optional feature

proposals/mystic.md
Kelmat-ai: Good shout, I've added that in there under "Maintenance & Ownership".

In essence: vault standard is OSS, lending market layer and curated lending vault are proprietary.

proposals/mystic.md
Kelmat-ai: Yeah, 100%.  It wasn't approved yet when we wrote this, but we're right on time to build the vault OSS standard on top of 112 straight away now. They'll go great together I think

proposals/mystic.md
Kelmat-ai: Yeah we neglected that bit in the proposal, sorry. Adding that under the same section we mentioned above, "Maintenance & Ownership".

TLDR we'll commit to actively maintain it for a minimum of 12 months after the end of milestone 2, funded by protocol operations. We won't request any funding for maintenance. Vault standard code will remain under MIT license so anyone can use it, Mystic code will not.

Thanks for all the questions @VinyoDA, appreciate you reading everything and going through the application!

Noosphere-314:
Hi Mystic team! I`m Mykola Ilchuk, Head of R&D, [Woof](https://woof.software/). We're preparing two Canton Dev Fund [proposals](https://forum.canton.network/t/woof-two-dev-fund-proposals-evm-side-compliance-risk-infrastructure-for-canton-seeking-feedback-champion/8796), and Mystic's curated lending markets are exactly the kind of protocol they're built for.

The main one is a Canton Risk Engine: open-source on-chain risk parameter management - a registry with governance + timelock, an off-chain simulator, and a public risk dashboard. Today every lending team on Canton has to build collateral-factor / cap / liquidation-threshold management from scratch; we're making that a shared public good, free to use.

I'll be straight about the ask and why: what actually moves proposals through the committee isn't general support but named intent - a team saying "we'd use this, specifically for X." If (and only if) that's true for you after a look at the write-up, a short comment on our forum thread naming which piece you'd consume would genuinely change the picture: https://forum.canton.network/t/woof-two-dev-fund-proposals-evm-side-compliance-risk-infrastructure-for-canton-seeking-feedback-champion/8796

Second, smaller one: our Compliance Middleware SDK (EVM-side, for Zenith) handles per-party compliance gating at the application level - positions and pool access, the things instrument-level registry controls can't reach. If your markets are permissioned on the EVM side, you may be a natural consumer there too - happy to walk through it.

Either way, we'd want your input on the parameter set so it fits real lending needs - that offer stands regardless of the above.

cashenLuke:
We at Cashen have also been considering building a product incorporating vaults, and through our research have talked to multiple teams that are either scoping projects that use vaults, or already built production vault code. 
All desire a Canton vault standard for the reasons known and stated in this PR.  
We believe that a collective narrow approach on vault standard requirements would yield the best result for the network.  We would also look to leverage existing code from the ecosystem where possible to reduce duplicate work and increase speed to production.  

Given this proposal commingles a> vault standard with b> a lending product, is there scope to narrow it to only consider the vault standard and move quickly on that so app teams can all use it to build their products on a unified foundation?

Kelmat-ai:
As an update on this - we are now in talks with 19+ parties building with vaults on Canton Network. We are building a working doc as a basis for the standard and running it by a tigh-knit group of close parties who will co-author the standard with us, and then a larger group that can share their insights of what the standard should look like to fit the multiple use cases being built on Canton at the moment before anything. More on this soon

Noosphere-314:
On the vault standard question, and the narrowing @cashenLuke raised, one observation from the EVM side that may save the CIP some scope.

ERC-4626 works as widely as it does because it standardizes accounting and share semantics only: deposit, mint, withdraw, redeem, and the conversion between assets and shares. It deliberately says nothing about risk. That omission is the reason a vault standard and a lending market with opinions about risk can coexist without one constraining the other.

In the curated-vault designs that came later, MetaMorpho-style, the curator's surface sits outside the token standard: a per-market supply cap, an allocation list the curator can reorder freely, and a timelock with a guardian veto on the changes that raise risk. Caps and the timelock are not part of the share accounting. They are a separate governed layer the vault reads from. Keeping them out of the standard is what let many curators with very different risk appetites ship on one interface.

Applied here, that suggests a split. The CIP covers accounting, share semantics, the deposit and withdrawal lifecycle, and the events an integrator needs to reconstruct vault state. Curator-set limits, the parameter surface, and the path by which a curator changes them stay outside it. A team that wants a curator with unilateral authority and a team that wants a timelock and a veto can then both conform to the same standard, which is what makes it adoptable across the ecosystem rather than only inside one protocol.

One question, since you have thought about this longer than we have: for the isolated vaults, do you expect the curator's parameter changes to take effect immediately, or do you plan a delay with an exit window for LPs? That answer tends to determine how much of the parameter surface can live inside the standard and how much has to sit next to it.

Kelmat-ai:
@Noosphere-314 We agree on the split, and that's our proposed design for the CIP too. The standard covers share accounting, the deposit and redemption lifecycle, and the events an integrator needs to reconstruct vault state. Valuation, strategy, fees, access and roles stay with the implementer, for the reason you give: a team that wants a curator with unilateral authority and a team that wants a timelock and a veto should both conform.

On our own implementation, we will be following Morpho curated vaults. Changes that raise risk go behind a timelock, changes that lower it take effect immediately, and the pending change is visible for the duration so an LP who doesn't like it can exit before it lands. The timelock length will be set per vault rather than fixed.

Worth saying that's an implementation choice rather than something the CIP will enforce. A vault with no timelock at all would still be conformant.

Kelmat-ai:
More broadly, and to keep this PR inline with CIP-100, here are the RFP's under which this proposal sits:
- RFP 13 "Payments & DeFi", which specifically mentions standards for DeFi - exactly what we're doing here;
- RFP 12 "RWA Standards", which openly calls for open standards around issuance and RWA workflows, which is also exactly, pretty much to a letter, what we're doing here with the introduction of a vault standard.

Think it's pretty clear we should assess this PR specifically as the introduction of the vault standard and nothing else @hythloda . Ok to keep as-is, or would you want us to introduce a new one?

Vault working doc in full flow now, we're working on it closely with @cashenLuke and the Cashen team and will soon share it for feedback with a dozen+ teams, including Obsidian , the Foundation itself and many more key players on Canton.

cashenLuke:
Thanks for the update to scope and plans @Kelmat-ai 
If acceptable for CF, I'd suggest revising this one (to focus just on the standard) to keep the comments and audit trail alive

Noosphere-314:
@Kelmat-ai, thanks for clarifying both the CIP boundary and your implementation choices. Keeping the core compatible with different risk policies and governance models is the right call.

If it is useful, we could contribute a small set of integration review cases for implementations that do timelock parameter changes. They would sit in implementation guidance or an optional extension rather than in the core, so a vault with no timelock stays conformant.

The first case is a change whose delay has elapsed but for which no follow-up transaction has run. Implementations differ here in a way integrators feel. In our Daml prototype the effective value is derived from the scheduled activation time, so a reader that takes only the stored base value can return a limit that is no longer in force. An implementation that requires an explicit acceptance transaction behaves the other way round. Neither is wrong, but an integrator has to know which one they are reading.

The second is the notice period itself. Two things have to hold for it to do its job. The intended LP or monitoring party needs a way to obtain the pending change and its activation time, which on Canton is a question of how that information is exposed rather than something that follows automatically. And being able to request redemption is not the same as being able to complete an exit before the change lands, which depends on available liquidity and, where a vault supports asynchronous redemption, on how requests already in flight are treated. How a pending parameter change interacts with a redemption request that is already open looks like a reviewable case to us.

We have an executable Daml prototype with tests around scheduled changes, grouped parameter updates and consumer reads. It is not a deployed product, but the review cases would come from that work rather than from theory.

Could you include us in the wider feedback round you are preparing with @cashenLuke and the Cashen team?

PR #51, opened 2026-03-05
Add Canton Participant Event Gateway proposal

illegalcall:
## Development Fund Proposal Submission

**Proposal file:**
`/proposals/canton-participant-event-gateway.md`

---

## Summary

This PR has been reframed from Canton Event Indexer into **Canton Participant Event Gateway**: an open-source, participant-local application event delivery service for Canton dApps.

The revised proposal makes one clear architectural decision: **direct Ledger API ingestion is the canonical source, and the funded core works end-to-end without PQS**. PQS and Canton Index are treated as complementary infrastructure, not required dependencies.

The gateway focuses on the app integration layer that teams repeatedly rebuild:

- Daml `interfaces` as first-class event selectors, with template selectors as explicit opt-in/fallback.
- Participant-local Ledger API ingestion with durable checkpointing.
- REST/OpenAPI event history and replay.
- Reliable webhooks with HMAC signatures, retries, dead-letter queue, and replay.
- Party-scoped JWT/OIDC authorization.
- Docker/Helm packaging and reference integrations.
- 25% of the total request held for adoption validation.

---

## Checklist

- [x] Proposal file added under `/proposals/`
- [x] Milestones and funding amounts defined
- [x] Acceptance criteria included
- [x] Alignment with Canton priorities described

---

## Notes for Reviewers

- Funding request remains **400,000 CC**, but the payment structure is now **120,000 / 120,000 / 60,000 / 100,000 CC**, with the final **100,000 CC** tied to external adoption validation.
- The proposal is no longer positioned as a PQS layer, PQS alternative, general query engine, GraphQL system, or typed analytical projection system.
- Requested primary SIG label: `dapp-integration`.
- `canton-apis` review may also be useful because the deliverable includes REST/OpenAPI and webhook APIs, but the proposal is intentionally scoped as application integration rather than base indexing/query infrastructure.
- Champion is still needed.

PR #113, opened 2026-03-20
Proposal: Data Standard

Desmonty-Kaiko:
Introduced the Kaiko Data Standard to provide a unified interface for financial data consumption on the Canton network, reducing vendor lock-in and integration complexity.

**Proposal file:**  
`/proposals/kaiko-data-standard.md`

---

## Summary
The Kaiko Data Standard defines a provider-agnostic DAML interface package for publishing and consuming on-ledger data (e.g., prices, reference rates, market indicators) on Canton. It eliminates bespoke bilateral integrations between oracle providers and application consumers, enabling any Canton application to switch or add data providers without DAML rewrites. The Standard is already running in production with institutional clients (Broadridge, Bitsafe, NCFX) and this proposal funds its formal publication into Splice, along with ongoing maintenance, documentation, and adoption incentives.

---

## Checklist

---

## Notes for Reviewers

### DIFF
--- a/proposals/kaiko-data-standard.md
+++ b/proposals/kaiko-data-standard.md
@@ -1,0 +1,220 @@
+# Kaiko Data Standard
+
+## Abstract
+
+Financial applications on Canton need reliable, real-time data: asset prices, reference rates, market indicators, and more. Today, every application that wants to consume this data has to build a custom integration with each data provider it uses. If it wants to switch providers, or add a second one, it has to rebuild that integration from scratch. This creates unnecessary complexity, slows down development, and locks applications into specific vendors.
+
+The Kaiko Data Standard fixes this by establishing a shared way for data providers to publish data on Canton and for applications to read it. Think of it like a common plug standard: once it exists, any device works with any socket. Any oracle provider that implements the Standard can be consumed by any Canton application without additional integration work, and any application built on the Standard can switch providers or add new ones without changing its code.
+
+This is not a theoretical proposal. Kaiko has already built and deployed this Standard in its production oracle, and it is actively being used by institutional clients today. It was developed in close collaboration with Digital Asset and has drawn adoption interest from DRW and Cumberland. The grant would fund the formal publication of the Standard into [Splice](https://github.com/hyperledger-labs/splice) (the Canton open-source repository), making it a true ecosystem-wide common good, along with the ongoing support and maintenance needed to keep it useful over time.
+
+For the Canton ecosystem, this means faster application development, more competition among data providers, and a stronger, more interconnected network overall.
+
+## Specification
+
+### 1. Objective
+
+**Problem:** Today, consuming on-ledger data creates tight coupling between application providers (consumers) and specific oracle implementations (producers). This coupling generates:
+
+* **Vendor lock-in:** switching oracle providers or delivery modes requires DAML rewrites  
+* **High adoption friction:** every new integration is a bespoke project  
+* **Slower innovation:** application teams hesitate to adopt new oracle delivery models due to interface breakage risk  
+* **Fragmented observability:** normalizing on-ledger metrics and telemetry across providers is difficult
+
+Every one of these problems slows down Canton adoption. New application teams spend time on integration plumbing rather than building their product. Existing applications are reluctant to add data feeds or switch providers because the cost is too high. The ecosystem as a whole remains more fragmented than it needs to be.
+
+**Intended outcome:** A formally specified and publicly available DAML interface, published in Splice, that any oracle provider can implement and any Canton application can consume. Application providers integrate once to the Data Standard and can thereafter switch oracle providers or delivery modes without changing their DAML application logic. The result is a more open, competitive, and composable data layer for the entire Canton network.
+
+### 2. What Already Exists vs. What Is Net-New
+
+**What already exists in Splice and adjacent efforts**
+
+Splice already defines a pattern for shared DAML interfaces. Existing examples include `splice-api-featured-app-v1`, which introduces the `FeaturedAppRight` and `FeaturedAppActivityMarker` interfaces. These establish that Splice supports modular, multi-implementer interface packages and that Canton applications can be built against stable, governance-neutral contracts. This architectural model is not new.
+
+There is no existing data layer standard in Splice. No DAML interface package currently defines how on-ledger data points (prices, reference rates, market indicators, or any structured financial data) should be published, versioned, or consumed in a provider-agnostic way. Each oracle integration on Canton today is bespoke and bilateral.
+
+**What the Kaiko Data Standard introduces**
+
+* A dedicated DAML interface package for the data layer, specifically covering on-ledger data point publication and consumption. This does not exist in Splice today.  
+* A reference implementation with a validated producer (oracle) and consumer (application) example, grounded in a live production deployment, not a prototype.  
+* Developer documentation specifically for oracle producers and application consumers in the context of the data layer.
+
+The Kaiko Data Standard deliberately mirrors the architectural pattern already established in Splice rather than inventing a new one. The contribution is a new application of a proven model to a domain (data) that is currently unaddressed.
+
+### 3. Implementation Mechanics
+
+The Kaiko Data Standard is implemented as a DAML interface package, following the same architectural pattern already established in Splice (e.g., `splice-api-featured-app-v1`). It defines:
+
+* A stable **data contract interface** for published data points on Canton  
+* A **versioned data model** that producers can extend and version independently (e.g., `DataPointV1`, `DataPointV2`), without breaking downstream consumers  
+* The **Standard itself** follows a named versioning scheme (`v1`, `v2`, etc.) to allow it to evolve over time without breaking existing implementations
+
+**Implementation workflow:**
+
+1. Finalize and formally specify the DAML interface definition  
+2. Implement and test the code against real oracle data pipelines (leveraging Kaiko's existing production deployment as a reference)  
+3. Conduct alignment sessions with Digital Asset and other data consumers to validate the design and gather implementation feedback  
+4. Submit the DAML package to the Splice repository via Pull Request, pending Digital Asset's review and approval  
+5. Publish reference documentation (developer guide for producers and consumers) on Kaiko's website and coordinate with the Canton ecosystem
+
+**Technologies:** DAML, Canton ledger, Splice (Hyperledger Labs), Kaiko's existing oracle infrastructure.
+
+**Operational approach:** Post-publication, the Standard is governed by Splice's open-source contribution process. Any party, whether oracle providers, application developers, or ecosystem participants, may propose changes, improvements, or new versions via GitHub Pull Requests. Kaiko does not retain unilateral control over the Standard post-merge. Kaiko commits to providing ongoing support, maintenance, and regular office hours to assist adopters, answer technical questions, and facilitate future contributions. This is a sustained commitment that requires dedicated resources beyond the initial build.
+
+### 4. Long-Term Ownership and Maintenance Post-Merge
+
+Once merged into Splice, the Standard is governed entirely by Splice's open-source contribution process under Hyperledger Labs. Kaiko does not retain unilateral control over the Standard, its roadmap, or its versioning decisions. Any party, whether oracle providers, application developers, or other ecosystem participants, may propose changes, new versions, or improvements via standard GitHub Pull Requests, subject to community review.
+
+Kaiko does not step back after the merge. Kaiko commits to the following ongoing responsibilities, funded in part by this grant:
+
+* **Ongoing maintenance:** Kaiko will respond to issues filed against the Standard on GitHub, submit bugfixes and clarifications, and maintain compatibility of its own production oracle with the published Standard.  
+* **Community Q\&A and issue resolution:** Kaiko will monitor and respond to community questions in relevant channels (e.g., Discord, GitHub Discussions, gsf-outreach) with a target response time of 5 business days.
+
+**Distinction from Kaiko's commercial product:** The Standard is deliberately decoupled from Kaiko's proprietary oracle product. Kaiko's commercial oracle implements the Standard, but the Standard itself does not depend on or require Kaiko's oracle. Any provider can implement it independently.
+
+### 5. Architectural Alignment
+
+The Kaiko Data Standard is a natural extension of Canton's interface-based composability model. Splice already establishes the pattern of defining DAML interfaces that multiple parties implement independently (e.g., `FeaturedAppRight`, `FeaturedAppActivityMarker`). The Data Standard applies this exact model to the data layer:
+
+* **Protocol-level standardization:** defines a shared language for on-ledger data points, reducing fragmentation across oracle implementations  
+* **Composability:** multiple applications can consume and re-use the same published data contracts, enabling shared infrastructure such as indexers, dashboards, and analytics tools  
+* **Openness:** by living in Splice (Hyperledger Labs), the Standard inherits community governance and is not controlled by any single vendor  
+* **Ecosystem resilience:** application providers can add secondary data sources or switch providers without rewrites, improving the long-term resilience of Canton-based applications  
+* **Alignment with ecosystem goals:** the Standard directly advances Canton's stated priorities of adoption, composability, and openness
+
+The ecosystem-level impact is concrete. A standardized data layer means:
+
+* New applications can go live faster because the data integration layer is already solved  
+* More oracle providers can enter the Canton ecosystem without needing bilateral agreements with every application  
+* The Canton network becomes more attractive to institutional participants who expect open, interoperable infrastructure as a baseline
+
+Beyond oracle data, the Standard also provides a foundation for normalizing application observability. Application providers can publish standardized metrics and events using the same interface, enabling shared dashboards and ecosystem-wide transparency.
+
+## Milestones and Deliverables
+
+### Milestone 1: Design, Community Alignment & Initial Pull Request
+
+* **Estimated Delivery:** 6 weeks from grant approval  
+* **Funding:** 300,000 CC upon committee acceptance  
+* **Focus:** Finalize the DAML interface specification, align with Digital Asset and other data consumers, and submit the initial Pull Request to the Splice repository
+
+**Deliverables / Value Metrics:**
+
+* Finalized DAML interface specification, including versioning strategy (data model versioning and Standard versioning)  
+* Documented alignment sessions with Digital Asset and at least 2 oracle providers (e.g., DRW, Cumberland)  
+* Reference implementation: at least one producer (oracle) and one consumer (application) example, validated against Kaiko's live oracle deployment (Broadridge, Bitsafe, NCFX)  
+* Pull Request publicly submitted to [hyperledger-labs/splice](https://github.com/hyperledger-labs/splice)
+
+**Ecosystem value:** The PR submission date and the number of external parties who have reviewed and provided documented feedback on the specification prior to submission. A Standard that reflects multi-party input at draft stage is significantly more likely to be adopted broadly.
+
+### Milestone 2: Merge Request Approved & Data Standard Deployed
+
+* **Estimated Delivery:** 10 weeks from grant approval  
+* **Funding:** 400,000 CC upon committee acceptance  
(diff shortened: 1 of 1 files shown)


amerr1man:
Given the likelihood of many oracle providers in the future, developing a standard oracle interface for Canton seems worthwhile.  Notably, Ethereum has implement a similar, simple oracle interface in https://eips.ethereum.org/EIPS/eip-7726 that is used by various applications including Euler.  I would love to hear other opinions but from my perspective this proposal seems ready to move forward for an official decision

mgaare:
I believe this probably needs to include a CIP.

mgaare:
It would be helpful if this proposal could include some more specific thoughts on what the design, enabled workflows, and Daml interfaces might look like.

mgaare:
Overall something like this is needed, and as you mentioned in the proposal, this is something that we at Denex/Cumberland are working with you on.

Desmonty-Kaiko:
Hello @mgaare, thanks for the review. We’ll update the proposal to address all three points:

1. CIP in Milestone 1: We’ll formally commit to drafting and submitting a CIP as a deliverable, ensuring the standard goes through proper community review.
2. Design and workflow details: We’ll expand the documentation to cover design principles, key workflows, and how the Daml interfaces fit together.
3. Example Daml interface: We’ll attach a concrete example to make the technical direction easier to evaluate.

We’ll also revise the Milestone 1 funding amount upward to reflect the additional scope that comes with drafting and shepherding a CIP through approval.

proposals/kaiko-data-standard.md
```diff
@@ -0,0 +1,296 @@
...
+-- example, BTC/USD at 45000.00 with `decimals = 8` is encoded as
+-- `price = 4500000000000`.
+data Quote = Quote
+  with
+    feedId : Text
+      -- ^ Stable identifier for the price feed (e.g. "BTC/USD", "ETH/EUR").
+    price : Int
+      -- ^ Integer-encoded price. `Int` is 64-bit signed.
+    decimals : Int
+      -- ^ Number of decimal places applied to `price`.
```
meiersi-da, line 94: Why not use Daml's `Decimal`, which uses 128-bit integers internally; and thus has a larger range than the 64-bit `Int`?

proposals/kaiko-data-standard.md
```diff
@@ -0,0 +1,296 @@
...
+
+-- | The strict payload returned by a `PublishedQuote` instance.
+--
+-- The real-world price is reconstructed as `price / 10 ^ decimals`. For
+-- example, BTC/USD at 45000.00 with `decimals = 8` is encoded as
+-- `price = 4500000000000`.
+data Quote = Quote
+  with
+    feedId : Text
+      -- ^ Stable identifier for the price feed (e.g. "BTC/USD", "ETH/EUR").
```
meiersi-da, line 90: I assume these are unique per `distributor` correct? 

Would it make sense to structure feed-ids such that some of them are portable across providers?

proposals/kaiko-data-standard.md
```diff
@@ -0,0 +1,296 @@
...
+
+-- | The standard interface that any quote producer must implement.
+--
+-- Consumers depend on this module only; they never need to know which
+-- concrete template produced the quote.
+interface PublishedQuote where
+  viewtype PublishedQuoteView
+
+  fetchValue : Quote
+    -- ^ Return the strict `Quote` payload. MUST equal `(view this).quote`.
```
meiersi-da, line 124: I believe you'll want a nonconsuming choice with a flexible controller akin to the `TransferFactory_PublicFetch` choice here: https://github.com/canton-network/splice/blob/1100d66a4da1121517ba6019ebf49f394009dee3/token-standard/splice-api-token-allocation-instruction-v1/daml/Splice/Api/Token/AllocationInstructionV1.daml#L130-L142

Otherwise, only stakeholders of the quote contract can read it, which makes it impossible to optimize quotes by publishing them only once on-ledger and then sharing them via explicit disclosure.

proposals/kaiko-data-standard.md
```diff
@@ -0,0 +1,296 @@
...
+- **Quote**: A specialized interface for quote price, one of the most common data structures required by most DeFi applications. This interface can be used to simplify implementation when it is known that only the price and timestamp are required.
+
+Both interfaces are used as follow:
+
+1. The Data Provider or Oracle creates a Daml template that implements the relevant interface from the Data Standard.  
+2. The Data Consumer creates a Daml template for their application and uses the Data Standard in place of the Oracle data point. The Data Consumer can implement the logic to fetch the price from any Daml template that implements the relevant interface from the Data Standard.  
+3. The Data Consumer gets access to a data point contract published by the oracle through any data delivery mechanism (push, pull, request-response, or any other), and uses the data point contract into their application workflow.
```
meiersi-da, line 61: I'd have expected that at least one of these options is standardized as well, as otherwise apps still have to do one-off integrations for getting the data contracts from different providers. Why is that not required?

proposals/kaiko-data-standard.md
```diff
@@ -0,0 +1,296 @@
...
+      -- ^ Ledger time at which the quote was published on-chain.
+    quote : Quote
+      -- ^ The strict, typed quote payload.
+  deriving (Eq, Show)
+
+-- | The standard interface that any quote producer must implement.
+--
+-- Consumers depend on this module only; they never need to know which
+-- concrete template produced the quote.
+interface PublishedQuote where
```
meiersi-da, line 120: There's an alternative design that uses less traffic: publish quotes together with a signature of the distributors and have users verify the signature as part of the Daml transaction transaction that uses the quote. 

I'd suggest standardizing both options, as they have different properties wrt revocation. The signature approach works a bit like JWT access tokens: no revocation, but short enough life-times that the damage from a wrong quote is limited. For high-stakes quotes, having the distributor check activeness of the contract and thus provide instant propagation of corrections seems like the better option.

meiersi-da (commented):
Thanks. Some comments from a first read.

proposals/kaiko-data-standard.md
Desmonty-Kaiko: Thanks for the feedback.
The price as Int with additional decimals field is based on existing format.
Considering Daml's `Decimal` type, I agree it is wiser to use it.

This interface serves as an example at this stage and is not optimized yet.
The interface will be clearly defined during the CIP creation process and become final when merged in splice.
At this stage, we only wanted to demonstrate the interface. We will also add usage examples that we didn't include here for the sack of clarity.

proposals/kaiko-data-standard.md
Desmonty-Kaiko: Uniqueness would not be enforced.
We plan for `feedId` to be informational and different distributor are free to use similar feedId (e.g., "BTC/USD") even when different distributor may use different calculation methodology to define their respective feeds.

proposals/kaiko-data-standard.md
Desmonty-Kaiko: Thanks! It should at the very least be nonconsuming.
Regarding the flexible controller, that could depend on use-cases, since some data provider prefer to control who can use their data on-chain. Nonetheless, I believe that the most frequent situation will be that having a flexible controller is useful.
I will add the flexible controller to the example so it commit to the general case and keep more specialized interfaces for the actual implementation the Data Standard.

proposals/kaiko-data-standard.md
Desmonty-Kaiko: You're right that switching providers under the current proposal still requires integration changes outside the app's DAML templates.

The Data Standard's specific purpose is narrower: prevent dependencies from an app's DAML templates on a specific oracle's templates. Decoupling the contract type via a shared interface is what lets apps swap providers without DAML rewrites, and it also enables multi-sourcing (running multiple providers in parallel for redundancy or arbitration).

Standardizing a delivery mechanism is a fair ask but a much larger problem, and the difficulty is that different commercial models naturally fit different on-chain patterns: per-query payment fits Request-Response, subscription fits Push with periodic deactivation, and so on. Forcing one delivery interface onto every provider would either constrain commercial models or push providers to bolt their licensing logic onto a shape that doesn't fit. The two standardizations (data shape and delivery mechanism) can be done separately, which is why prefer to focus on the Data Standard first.

Kaiko plans to propose a follow-up CIP for the on-chain delivery patterns of an Oracle, targeted for Q3.

meiersi-da: > Standardizing a delivery mechanism is a fair ask but a much larger problem, and the difficulty is that different commercial models naturally fit different on-chain patterns: per-query payment fits Request-Response, subscription fits Push with periodic deactivation, and so on. Forcing one delivery interface onto every provider would either constrain commercial models or push providers to bolt their licensing logic onto a shape that doesn't fit. The two standardizations (data shape and delivery mechanism) can be done separately, which is why prefer to focus on the Data Standard first.

Fair point. What do you think about the middle ground where:

1. the standard defines at least one delivery method; e.g., the Oracle delivery method.
2. the standard calls out that providers are free to use alternative delivery methods that fit their licensing models better, and that we expect these delivery methods to be standardized as well in future CIPs.

I feel without at least one delivery method the standard does not solve its goal of making the switch between data providers a no-op.

I also suspect that investigating the delivery methods will influence the design of the Daml interfaces. So it seems valuable to me to at least have an overview of the different methods. What I'd suggest for that is to add a "Usecase" section, where the different licensing use cases are described, and it is shown how they fit with the proposed APIs.

proposals/kaiko-data-standard.md
```diff
@@ -0,0 +1,296 @@
...
+
+* **Vendor lock-in:** switching oracle providers or delivery modes requires DAML rewrites  
+* **High adoption friction:** every new integration is a bespoke project  
+* **Slower innovation:** application teams hesitate to adopt new oracle delivery models due to interface breakage risk  
+* **Fragmented observability:** normalizing on-ledger metrics and telemetry across providers is difficult
+
+Every one of these problems slows down Canton adoption. New application teams spend time on integration plumbing rather than building their product. Existing applications are reluctant to add data feeds or switch providers because the cost is too high. The ecosystem as a whole remains more fragmented than it needs to be.
+
+**Intended outcome:** A formally specified and publicly available DAML interface, published in Splice, that any oracle provider can implement and any Canton application can consume. Application providers integrate once to the Data Standard and can thereafter switch oracle providers or delivery modes without changing their DAML application logic. The result is a more open, competitive, and composable data layer for the entire Canton network.
```
daravep, line 26: Thank you for this proposal. I do think that standardization around Oracle data is indeed a highly valuable contribution. The mechanisms of introducing this using a CIP will also provide the subsequent process, where the ecosystem can contribute and discuss.

Is it correct that the commercial model for data providers will be to benefit from featured app rewards whenever the Oracle data is being referenced? 

A minor extension of the acceptance criteria could be to use the data standard in the splice models as part of the adoption / validation effort. If we have the standard, we might also start its proliferation by using it in the core models that all validators install.

proposals/kaiko-data-standard.md
meiersi-da: The choice implementation can still limit what `actors` are allowed. We've made good experiences with using the `actors` pattern uniformly in the token standard V2 interfaces:
* example: https://github.com/canton-network/splice/blob/30c75481b46088792b05b2f9d910e89dfbe08040/token-standard/splice-api-token-allocation-v2/daml/Splice/Api/Token/AllocationV2.daml#L316 (also see the `availableActions` in the `AllocationView` to communicate who can call the choice.)
* context: https://lists.sync.global/g/cip-discuss/topic/cip_0112_token_standard_v2/118630150

proposals/kaiko-data-standard.md
meiersi-da: OK, it sounds fine for every distributor to manage their `feedId`s by themselves. However, for them to not guarantee that a `feedId` always refers to the same feed seems problematic. That's what I meant with uniqueness per provider.

proposals/kaiko-data-standard.md
Desmonty-Kaiko: Thanks for raising this. The two patterns have distinct properties regarding revocation: contract-based provides instant correction propagation, while signature-based offers traffic efficiency with bounded-damage TTLs. I agree the Data Standard should cover both, so that the use case can dictate the choice rather than the available primitive.

I will add a second interface alongside PublishedQuote in the CIP and document the revocation tradeoffs so consumers can pick deliberately based on their data product.

proposals/kaiko-data-standard.md
Desmonty-Kaiko: > Is it correct that the commercial model for data providers will be to benefit from featured app rewards whenever the Oracle data is being referenced?

Yes, partially. Under CIP-0104, exercising a nonconsuming fetch choice on the PublishedQuote entitles the oracle to a share of the app rewards for that envelope, and the publication transaction is also rewarded.
This does not cover the entire commercial model. Oracles remain free to layer additional commercial logic such as subscription or pay-per-use on top.

> A minor extension of the acceptance criteria could be to use the data standard in the splice models as part of the adoption / validation effort. If we have the standard, we might also start its proliferation by using it in the core models that all validators install.

Thanks for the suggestion. The Daml interface will be published in splice so all validators have it available by default. Adoption in specific splice core models would be valuable, but depends on the maintainers of those models, so I would prefer not to include it as acceptance criteria here. Happy to engage with the relevant teams as a follow-up.

meiersi-da: > Thanks for the suggestion. The Daml interface will be published in splice so all validators have it available by default. Adoption in specific splice core models would be valuable, but depends on the maintainers of those models, so I would prefer not to include it as acceptance criteria here. Happy to engage with the relevant teams as a follow-up.

There's the CC/USD price feed that is consumed by the SV app to publish an on-ledger price vote. It would be great to convert that feed to support the new data standard as part of doing the data standard.

This way we have at least one realistic test case for the new APIs. wdyt?

proposals/kaiko-data-standard.md
Desmonty-Kaiko: I just saw that Contract Keys are coming back to Canton ‽ This is really awesome.
Then, it would make sense to have a key on (distributor, feedId) though I'm not sure how to enforce it at the interface level.
I'll note this point and ensure it is raised during the implementation of the Data Standard.

meiersi-da: Yeah, there are no interface keys at this point of time.

And I was only expecting that distributors manage these id's such that consumer can them to be unique per `distributor` party. How the distributors do that is up to them. Doesn't have to be on-ledger.

Desmonty-Kaiko: We can leave a comment for distributors in the Data Standard interface that says they are expected to maintain uniqueness somehow for feedIds they produce.

proposals/kaiko-data-standard.md
Desmonty-Kaiko: > I feel without at least one delivery method the standard does not solve its goal of making the switch between data providers a no-op.

The goal is narrower than no-op switching; it is rather to remove the dependency between an Application's Daml templates and a specific Oracle's Daml templates. Today, app providers compile their templates against the oracle's codebase, so switching providers forces a full application migration and breaks running contracts. The Data Standard solves that. Switching providers will still require integration work, since providers legitimately differ in data quality checks and commercial models.

> I also suspect that investigating the delivery methods will influence the design of the Daml interfaces. So it seems valuable to me to at least have an overview of the different methods. What I'd suggest for that is to add a "Usecase" section, where the different licensing use cases are described, and it is shown how they fit with the proposed APIs.

Kaiko will add a Use Case section to the CIP including a catalogue of reference implementations covering distinct distribution models (at least contract-based and signature-based), each showing how the Data Standard interfaces apply. The catalogue draws on Kaiko's existing oracle implementations in production, providing concrete starting points anyone can adapt to their needs. This fits within the existing milestones, and I can update the Acceptance Criteria accordingly.

I suggest that standardizing the delivery mechanism itself remains out of scope, to keep this proposal focused on the data standard. A follow-up CIP can address it, drawing on the patterns already validated by production deployments.

proposals/kaiko-data-standard.md
Desmonty-Kaiko: Yes, this is a good test case. Kaiko will provide a reference implementation of the Data Standard for the AmuletConversionRateFeed as part of Milestone 3, and collaborate with the splice maintainers on integration into the SV app. I will update the Acceptance Criteria accordingly.

hythloda (approved):

illegalcall:
Hi @Desmonty-Kaiko, targeted design-partner question from #51, not a request to change this approved proposal.

I revised #51 from an indexer/PQS-adjacent proposal into **Canton Participant Event Gateway**: direct Ledger API ingestion, interface-aware event subscriptions, REST replay/event history, reliable webhooks, delivery audit, and party-scoped auth.

For the Data Standard / oracle workflow, would interface-aware event subscriptions and replayable webhook delivery be useful for consumers watching oracle publication/update events or building off-ledger risk/monitoring workflows around the standard?

If you have a quick yes/no, or a concrete requirement that would make this useful/not useful, a comment on #51 would help validate the revised interface-first scope.

pedrodneves:
### Milestone 1: Design, Community Alignment & CIP

**Deliverables / Value Metrics:**

* Finalized DAML interface specification, including versioning strategy (data model versioning and Standard versioning)  
* Documented alignment sessions with Cumberland, Digital Asset and at least 2 oracle providers.  
* Reference implementation: at least one producer (oracle) and one consumer (application) example, validated against Kaiko's live oracle deployment (Broadridge, Bitsafe, NCFX)  
* CIP submitted and approved

**Ecosystem value:** The CIP approval date and the number of external parties who have reviewed and provided documented feedback on the specification prior to submission. A Standard that reflects multi-party input at draft stage is significantly more likely to be adopted broadly.


Estimated Delivery | Funding | Focus
-- | -- | --
8 weeks from grant approval | 900,000 CC upon committee acceptance | Finalize the DAML interface specification, align with Cumberland, Digital Asset, and data consumers, and drive a CIP from drafting through submission and approval.

pedrodneves:
### Milestone 2: Merge Request Approved & Data Standard Deployed

**Deliverables / Value Metrics:**

* Complete DAML package implementation with a passing test suite covering producer and consumer interface conformance  
* All review comments addressed; PR approved by Digital Asset to publish Data Standard in splice.  
* DAML package successfully merged into Splice and available to all Canton participants  
* Developer documentation published on Kaiko's website (integration guide for oracle producers and application consumers)

**Ecosystem value:** Public documentation and open-source codebase published in splice codebase.

Estimated Delivery | Funding | Focus
-- | -- | --
4 weeks from CIP approval (Milestone 1 completion) | 400,000 CC upon committee acceptance | Implement and test the DAML package, incorporate review feedback from Digital Asset and the Splice community, reach merge approval, and deploy the Standard in Splice, making it universally accessible to all Canton participants.

pedrodneves:
### Milestone 3: Ecosystem Outreach, Use Case Showcase & Adoption Incentive

**Deliverables / Value Metrics:**

* Public outreach campaign published on LinkedIn and via gsf-outreach channels, highlighting the Standard's availability and value  
* Optional collaboration with the Canton Marketing Committee on announcement coordination and amplification  
* Published use case showcase featuring at least 1 live user of the Data Standard (from existing users: Broadridge, Bitsafe, NCFX, or new adopters), illustrating a concrete integration and the interoperability benefits achieved  
* Per-client payment: 100,000 CC released upon committee acceptance that a client has successfully integrated and gone live on the Data Standard, up to 10 clients

**Ecosystem value:** Number of independent Canton applications consuming live data via the Standard. This directly measures whether the Standard has reduced integration friction at scale, not just whether it was published.

**Rationale for the per-client adoption incentive structure:** By publishing the Data Standard as a common good and making it freely available through Splice, Kaiko is deliberately reducing its own commercial stickiness. The Standard enables Kaiko's existing clients to more easily switch to an alternative data provider, as it removes proprietary API lock-in as a retention mechanism. This is a concrete commercial sacrifice made in the interest of ecosystem health. The per-client adoption payment compensates Kaiko for this accepted increase in churn risk, and aligns the Foundation's incentive with the Standard's actual adoption. The more ecosystem participants adopt it, the greater the common good delivered, and the greater the compensation.

Estimated Delivery | Funding | Focus
-- | -- | --
Up to 12 months from Milestone 2 completion | 100,000 CC per client that adopts and goes live on the Data Standard, up to a maximum of 10 clients (1,000,000 CC maximum) | Drive ecosystem-wide awareness and reward demonstrated real-world adoption of the Standard by Canton application providers.

PR #124, opened 2026-03-23
Proposal : Visibility regression proposal

deepthi-2531:

**Proposal file:**  
`/proposals/canton-visibility-regression-kit-for-jvm.md`

---

## Summary
This proposal requests funding for an open-source JVM testing toolkit that helps Canton teams detect privacy and visibility regressions before they reach production workflows. It delivers reusable visibility assertions, party-scoped query helpers, regression diffing, CI-grade reports, and multi-party reference scenarios that make Canton’s privacy guarantees easier to test repeatedly and correctly.

---

## Checklist

---

## Notes for Reviewers
This proposal is scoped as privacy-correctness testing infrastructure, not a workflow engine or formal verification system. The main value is giving teams a practical, repeatable way to verify who can and cannot see contracts across workflow transitions, and to catch observer leaks or unintended stakeholder-set drift early in CI.

### DIFF
--- a/proposals/canton-visibility-regression-kit-for-jvm.md
+++ b/proposals/canton-visibility-regression-kit-for-jvm.md
@@ -192,11 +192,11 @@ The toolkit remains usable after a second rerun or follow-up evaluation cycle an
 
 ## Potential Ecosystem Beneficiaries
 
-This proposal is intended as public-good testing infrastructure for the wider Canton ecosystem, and I have identified a few ecosystem teams that are well aligned with this kind of capability and have expressed interest in this kind of privacy-correctness tooling, including `Gateway`, `Lumens.fi`, and `BitDynamics AB`.
+This proposal is intended as public-good testing infrastructure for the wider Canton ecosystem. Early conversations indicate that this kind of privacy-correctness tooling is relevant to ecosystem teams building privacy-sensitive Canton applications, including teams such as `Gateway` `Lumens.fi` and `BitDynamics AB`
 
-The feature set addresses a common Canton-specific risk: multi-party workflows are easy to get functionally correct while still getting privacy boundaries wrong through observer leaks, stakeholder creep, or unintended visibility changes.
+The feature set addresses a common Canton-specific failure mode: multi-party workflows are often easy to make functionally correct while still getting privacy boundaries wrong through observer leaks, stakeholder creep, or unintended visibility changes.
 
-More broadly, this project is useful for teams building privacy-sensitive Canton applications who want practical CI-grade checks for party visibility, workflow transitions, and regression detection around access boundaries.
+More broadly, the project is useful for teams building wallet, gateway, trading, payment, and other multi-party Canton workflows who need practical CI-grade checks for party visibility, workflow transitions, and regression detection around access boundaries.
 
 
 ## Acceptance Criteria
--- a/proposals/canton-visibility-regression-kit-for-jvm.md
+++ b/proposals/canton-visibility-regression-kit-for-jvm.md
@@ -1,4 +1,4 @@
-## Development Fund Proposal: Canton Visibility Regression Kit for JVM
+## Development Fund Proposal: Canton Visibility Regression Kit for JVM and CI for all other languages
 
 - **Author:** Deepthi  
 - **Status:** Submitted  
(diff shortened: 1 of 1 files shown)


srikanth-bitdynamics:
@deepthi-2531  This is a good tool idea but can you not just limit this to Java and extend it to other tech stacks as well.

deepthi-2531:
@srikanth-bitdynamics  Thanks for the feedback I updated the proposal today to have a CI which can help apps on any tech stack to get regression analysis.

srikanth-bitdynamics:
Thanks. Can you add a bit on how upgrades are planned for this post release?

deepthi-2531:
Updated it @srikanth-bitdynamics

srikanth-bitdynamics:
Thanks. 

My Review:

This project addresses a high-value Canton-specific problem: privacy and visibility regressions are easy to introduce, difficult to catch with normal functional tests, and central to Canton’s core differentiator of controlled disclosure. The proposal is strong because it stays narrowly focused on that problem while delivering practical tooling teams can actually use in tests and CI.

The implementation is now credible end to end after feedback: JVM core, reporting, JUnit integration, CLI/CI path for non-JVM stacks, live Canton validation on Daml/Canton 3.4.x, and a documented OSS upgrade/compatibility policy. In my view, this makes it a solid piece of ecosystem infrastructure rather than just a collection of test helpers.

I recommend this proposal proceed to voting. 

cc: @hythloda

proposals/canton-visibility-regression-kit-for-jvm.md
```diff
@@ -0,0 +1,360 @@
...
+
+to determine whether a workflow reveals too much to the wrong parties.
+
+That gap is risky because privacy regressions are easy to introduce and expensive to detect late. A workflow can be functionally correct and still violate the privacy model it was supposed to preserve. As more Canton applications move from prototypes to real multi-party systems, the ecosystem needs shared infrastructure that makes privacy boundaries testable on every change.
+
+### 3. Implementation Mechanics
+
+The project will be delivered as:
+
+- a JVM testing library for Canton visibility assertions
```
daravep, line 71: Why do you think that this should be built in the JVM and not as part of the daml native tests (daml script)?  

Validating the visibility of the state after a step can be done in daml script as part of the daml tests using https://docs.digitalasset.com/build/3.4/reference/daml-script/api/Daml-Script.html#function-daml-script-internal-questions-query-query-55941

Validating the visibility of disclosed contracts during a transaction is indeed something that would be an extension to assert on, but then again, I would assume that extending submitAndResultTree https://docs.digitalasset.com/build/3.4/reference/daml-script/api/Daml-Script.html#function-daml-script-internal-questions-submit-submitresultandtree-13546 with witness information might be the more convenient approach.

I think that the UX is much better if the visibility assertions are close to the actual code and don't exist separately in a CI step which requires to maintain multiple data dependencies spread across multiple tools.

deepthi-2531: Hey @daravep,

My reasoning for choosing the JVM has two parts:

1. The Daml docs position Daml Script and native tests as useful for validating ledger-model fundamentals early, while application-level integration testing naturally moves toward the Ledger API. This project is intended to verify privacy outcomes at that integration layer: which parties can actually see which contracts before and after workflow transitions. A Canton application may be driven by Java, TypeScript, Python, Go, or multiple services together. A JVM core plus a CLI/CI layer makes the toolkit usable across those stacks, rather than limiting it to Daml-native test suites.

2. Java is still one of the most common enterprise application languages, so a JVM implementation gives a practical adoption path for many teams immediately. It also lets the project expose a reusable library for JVM teams while still supporting non-JVM teams through CLI-based CI integration.

More broadly, my goal here was to avoid making privacy regression testing tightly coupled to a single Daml package test environment. If this lived only as Daml-native tests, each team would still need to adapt it into its own application and CI setup. By placing it at the Ledger API / toolkit layer, it becomes reusable ecosystem infrastructure.

hythloda:
To strengthen the proposal, it would also be helpful to include clear adoption signals from the ecosystem. For example, identifying two to three Canton ecosystem teams who intend to use the tool would help demonstrate practical demand. Testing the framework against existing projects such as Splice or the Token Standard could also provide valuable validation.

Please add this additional information directly to the PR. Once the proposal has been updated with these details, the team is welcome to reopen it for further review.

PR #126, opened 2026-03-23
Proposal: Lag aware pruning coordinator proposal

srikanth-bitdynamics:

**Proposal file:**  
`/proposals/canton-lag-aware-synchronizer-pruning-coordinator.md`

---

## Summary
This proposal requests funding for an open-source operator toolkit for lag-aware, policy-driven synchronizer pruning coordination across sequencer, mediator, and BFT-orderer components. The project focuses on safer pruning workflows through unified status collection, dry-run planning, blocker diagnosis, policy guardrails, and operator runbooks, helping teams reduce storage sprawl and pruning-related operational risk without changing Canton protocol semantics.

---

## Checklist

---

## Notes for Reviewers
This proposal is scoped as operator tooling and reference infrastructure around existing Canton pruning workflows, not as a protocol change, new pruning algorithm, or consensus/storage-engine modification.

The main differentiators are:
- focused on diagnostics, dry-run planning, and policy-gated coordination rather than blind pruning automation
- designed to normalize pruning posture across sequencer, mediator, and BFT-orderer layers
- explicitly conservative under partial visibility or unsupported adapter surfaces
- includes reference scenarios for blocked pruning, partial BFT visibility, and force-prune impact preview

For v1, BFT-orderer support is intentionally limited to read-only and diagnostic-first coverage where supported. Full BFT execution parity is not required for milestone acceptance in the first release.

### DIFF
--- a/proposals/canton-lag-aware-synchronizer-pruning-coordinator.md
+++ b/proposals/canton-lag-aware-synchronizer-pruning-coordinator.md
@@ -1,0 +1,380 @@
+## Development Fund Proposal: Lag-Aware Synchronizer Pruning Coordinator
+
+**Author:** Srikanth (BitDynamics)  
+**Status:** Submitted  
+**Created:** 2026-03-22  
+
+---
+
+## Abstract
+
+This proposal requests funding for an open-source operator toolkit for **lag-aware, policy-driven synchronizer pruning coordination** across sequencer, mediator, and BFT-orderer components.
+
+Canton already provides pruning commands and scheduling capabilities, but safe pruning remains operationally complex. Sequencer pruning can be blocked by inactive clients, force-pruning can disable members, BFT-orderer retention requires operator judgment, and the relationship between acknowledgements, lagging members, and safe pruning windows is still too manual for many teams.
+
+The proposed project, **Lag-Aware Synchronizer Pruning Coordinator**, will provide:
+
+- a unified status collector for sequencer, mediator, and BFT-orderer pruning state
+- dry-run pruning plans with blocker diagnosis and impact analysis
+- a policy engine for coordinated pruning schedules and retention guardrails
+- audit-friendly reports, dashboards, and operator runbooks
+
+The goal is not to change Canton pruning semantics and not to automate destructive actions blindly. The primary value of the project is diagnostics, normalization, and operator decision support. Execution capabilities, if included, remain explicitly policy-gated and secondary to the dry-run and reporting path. The goal is to reduce operator error, storage sprawl, and recovery risk by making existing pruning workflows safer and easier to understand.
+
+---
+
+## Team Background
+
+### BitDynamics
+
+BitDynamics brings deep experience in building and operating blockchain infrastructure. The team has worked across Ethereum client infrastructure, validator operations, and production-grade hosting systems supporting validator infrastructure securing more than 2 billion AUD in assets. This background is directly relevant to building reliable, auditable, and security-conscious public infrastructure for a grants program. The team is also building actively on Canton.
+
+---
+
+## Specification
+
+### 1. Objective
+
+The objective is to reduce operational risk and cost around synchronizer pruning by publishing a concrete, open-source pruning coordination toolkit that helps operators:
+
+- understand why pruning is blocked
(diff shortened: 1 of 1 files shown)


LucasnFCS (commented):
This proposal is well positioned as operator tooling around existing Canton behavior, not as a protocol change, and that is the right framing. In Canton, pruning is not just a local housekeeping task. Safe pruning depends on the state and lag posture of multiple synchronizer-adjacent components, and operational risk increases quickly when sequencer, mediator, and ordering layers are not evaluated together. From that perspective, a lag-aware pruning coordinator is a relevant infrastructure contribution.

A strong point is the proposal’s emphasis on status collection, dry-run planning, blocker diagnosis, and policy guardrails instead of aggressive automation. That fits Canton well. In practice, operators need confidence that pruning will not create avoidable recovery risk or blind spots across components with different visibility and support surfaces. A tool that helps normalize pruning posture, explain why pruning is blocked, and preview the impact of force-prune decisions is more valuable than a system that simply triggers pruning faster.
Its important to note that this is critically affecting the operations of Super Validators it should be created by a team with deep connections to the SV experience.

The scope also feels appropriately conservative. Keeping this as a reference toolkit and workflow layer rather than touching Canton protocol semantics, storage internals, or consensus behavior keeps the implementation risk much lower. That is especially important in a system like Canton, where operational correctness across components matters more than adding clever but hard-to-trust automation.

The proposal is also aligned with a real operator need: as Canton environments grow, storage sprawl and retention management become increasingly operational rather than theoretical problems. A consistent pruning coordination layer could reduce manual checks, reduce the chance of unsafe sequencing of actions, and make runbooks more repeatable across teams.

The main thing I would want to see evaluated carefully is whether the proposal defines clear and actionable notions of lag, readiness, and blockers across the different component types. That is where the real value will come from. The output needs to be operationally meaningful, not just a collection of raw status signals. Similarly, milestone acceptance should depend on demonstrable operator workflows, such as handling blocked pruning, partial visibility, and dry-run review quality, rather than only on code delivery.

The v1 limitation around BFT-orderer support being diagnostic-first and read-only where supported is reasonable. Given the differences in deployment surfaces and visibility, that is a sensible first step. But it also means users should view v1 primarily as a safer coordination and diagnostics layer, not full end-to-end pruning parity across all layers.

Overall, this looks like a practical and technically sensible proposal for the Canton ecosystem. It addresses a real operational gap, stays conservative where it should, and avoids overreaching into protocol design. Its success will depend less on novelty and more on whether it produces trustworthy diagnostics, useful dry-run planning, and operator workflows that teams would actually adopt in production.

Harsh-Zynk:
This looks like a practical and valuable proposal to us at Zynk.

From a Zynk perspective, lag-aware pruning coordination is the kind of operational gap that becomes very real once teams start running production-style Canton environments. Without shared tooling, different operators will likely end up reinventing similar safety checks and coordination logic themselves.

An open-source reference implementation here would be useful. We would be very much interested to integrate and use this once implemented

ghost:
We’d be interested in this tool to improve our current workflow. We'd be happy to try it out.

slugmann321:
At H2O Nodes we'd also be interested in testing this solution

ghost:
Gateway.FM supports this proposal as well and we are open to test it.

nogods404:
We're interested in this as it is a real problem for anyone currently running a validator.

Healthier pruning matters to teams building on Canton as well, and we are open to using it.

isegall-da:
The proposal focuses on pruning of sequencer, mediator and BFT-orderer, i.e pruning of Super Validators. On the global synchronizer, all Super Validators prune at a static pruning window of 30 days, and for many reasons, keeping that window constant is crucial, so there is no operator decision involved, and I therefore don't see what operator-supporting tooling is required.

The supportive comments above refer to validator pruning, which seems to completely be exempt from this proposal. If the author were to pivot the proposal into one for validator pruning, I'd urge them to make a clear statement on the current gap, with evidence that there is a problem to be solved. Pruning is supported in Canton participants, and it is aware of the lag, and will not prune any data that is unsafe to be pruned, so also for validators I am unclear what is the current gap in the technology that should be addressed.

PR #138, opened 2026-03-26
Proposal: Canton Network Indexer

zpoken:
## Summary

- **Canton Network Indexer** — three-tier indexing infrastructure for Canton: public network data (Scan API), party-scoped private contracts (Ledger API/PQS), and cross-party aggregation with RBAC
- Fills a gap no existing indexer (The Graph, SubQuery) can address due to Canton's sub-transaction privacy model
- Working PoC with 24/24 passing tests: party isolation, lifecycle queries, observer access, JWT auth
- 16-week roadmap, 3 milestones, 625,000 CC milestone-based funding

## Proof of Concept

PoC repo: https://github.com/zpoken/canton-network-indexer

## Team

[Zpoken](https://zpoken.io) — 50+ blockchain projects, $1.5B+ TVL shipped. Core contributors to Wormhole (ZK light clients), NEAR (ZK bridge), Polygon, and others.

## Test plan

- [x] PoC smoke tests pass (24/24)
- [x] Party isolation verified (Alice ≠ Bob)
- [x] JWT auth enforced at SQL level
- [x] Contract lifecycle queries (create → exercise → archive)
- [x] Observer access (regulator sees bonds)
- [x] Docker Compose reproducible deployment


proposals/canton-network-indexer.md
```diff
@@ -0,0 +1,427 @@
+# Proposal: Canton Network Indexer
+
+**Champion:** Canton Foundation
+
+## Objective and Scope
+
+We want to build a production indexer for Canton. Right now one doesn't exist, and it can't be solved by plugging in The Graph or SubQuery — Canton's sub-transaction privacy model breaks every assumption those tools make. There's no global state to scrape. Each participant node only sees contracts where its hosted parties are stakeholders, and everything else is encrypted.
```
hrischuk-da, line 7: I think there are actually two proposals here that need to be separated:  (1) a layer above PQS and (2) a global indexer.   By clubbing these two together, it weakens the proposal overall.  I would suggest dropping (2).

My feedback below is about (1) because this proposal can fill a gap that would benefit the community.

proposals/canton-network-indexer.md
```diff
@@ -0,0 +1,427 @@
...
+
+We've looked at how six other privacy chains handle this (Aztec, Aleo, Zcash, Namada, Secret Network, Mina). None of them are supported by any general-purpose indexer either. Every single one built custom infrastructure. Canton needs the same.
+
+The indexer is a three-tier system:
+- **Tier 1** indexes public network data (CC transfers, minting rounds, validators, governance) from the Scan API
```
hrischuk-da, line 12: There are many tools already available that summarize this data and have a head start.  I expect their APIs and offerings will mature quickly.  It isn't clear what additional value this brings and how the team would catch up.

proposals/canton-network-indexer.md
```diff
@@ -0,0 +1,427 @@
...
+- **Tier 2** indexes party-scoped private contracts (lifecycle events, ACS, audit trails) from the Ledger API via PQS
```
hrischuk-da, line 13: I applaud your use of PQS as the interface.

proposals/canton-network-indexer.md
```diff
@@ -0,0 +1,427 @@
...
+- **Tier 3** aggregates across parties within an organization, with RBAC for regulatory and audit views
```
hrischuk-da, line 14: This is where a lot of value can be added by closing known gaps.

proposals/canton-network-indexer.md
```diff
@@ -0,0 +1,427 @@
...
+
+Without this, every team building on Canton — tokenized assets, DvP settlement, fund admin, regulatory reporting — ends up rebuilding the same ingestion pipeline from scratch on top of PQS. That's wasted effort across the entire ecosystem.
```
hrischuk-da, line 16: This is not quite true.  Teams are happy to use the SQL API provided by PQS.  These integrations are usually fit-for-purpose for each application so the intersection of common use cases is important.  Supporting the following use cases is where the value would be and I would like to see the proposal focus on these:

- Providing JWT based access to PQS data.  This is something that can be reused and Validator-as-a-Service could leverage this.
- Pre-built, account like APIs that work with the token standard and other key CIPs.  This should help Ethereum developers adapt to our UTXO model.
- Aggregation across multiple PQS instances. Since there can be a PQS instance per validator it would be helpful to be able to  aggregate information (e.g., reporting) across multiple PQS instances.

This is not complete but should be representative.

proposals/canton-network-indexer.md
```diff
@@ -0,0 +1,427 @@
...
+
+**In scope:** public + private indexing, GraphQL/REST/SQL query layer, JWT party-scoped auth, Docker Compose and K8s deployment, Apache 2.0 open source.
```
hrischuk-da, line 18: More detail please.

proposals/canton-network-indexer.md
```diff
@@ -0,0 +1,427 @@
...
+
+**Out of scope:** custom Daml app development, managed hosting, changes to Canton internals, cross-chain indexing.
+
+## Technical Approach
+
+### Architecture
+
+```
+┌─────────────────────────────────────────────────────────────────┐
```
hrischuk-da, line 27: At the risk of being repetitive:

- Tier 1 is already supported by multiple tools
- Tier 2 is already supported by PQS
- Tier 3 is where I suggest you focus

proposals/canton-network-indexer.md
```diff
@@ -0,0 +1,427 @@
...
+Participant Node ──gRPC──> PQS (scribe.jar) ──> PostgreSQL (JSONB)
+                                                      │
+                                                  pg_notify
+                                                      │
+                                                Post-Processor ──> contract_state
+                                                                   party_activity
+                                                                   template mat.views
+```
+
+PQS handles the hard parts: gRPC streaming, offset tracking, Daml-LF serialization, crash recovery. We don't want to rewrite that. Instead we extend PQS output with a post-processor that populates indexer-specific tables (contract lifecycle state, per-party activity timelines, template-specific materialized views).
```
hrischuk-da, line 57: Please rewrite this to point out what value this is providing because PQS seems to be able to do all this.

proposals/canton-network-indexer.md
```diff
@@ -0,0 +1,427 @@
...
+
+### Consistency
+
+PQS watermark = consistency boundary. All data at or below the watermark is fully committed — no partial transactions visible. The indexer exposes watermark lag via health endpoint. Clients can query "as of offset N".
+
+Idempotent writes via `INSERT ... ON CONFLICT (offset, contract_id) DO NOTHING`. Post-processor resumes from last checkpoint on restart.
+
+### Security
+
+- JWT with `readAs`/`actAs` party claims at every API layer
```
hrischuk-da, line 79: How is this indexer JWT token's party claims synchronized with the validator's party claims?

proposals/canton-network-indexer.md
```diff
@@ -0,0 +1,427 @@
...
+- mTLS between indexer and participant node
+- Schema-per-party postgres isolation
+- RBAC matrix for Tier 3 cross-party access
```
hrischuk-da, line 82: More detail please

proposals/canton-network-indexer.md
```diff
@@ -0,0 +1,427 @@
...
+
+**PQS as foundation, not a custom gRPC client.** PQS is maintained by Digital Asset. It already deals with Ledger API evolution (services got removed between 3.3 and 3.5). Building a custom ingester means tracking all of that ourselves. Not worth it unless we're past 1000 tx/s.
+
+**Event sourcing.** The raw offset stream is the source of truth. Materialized views are projections you can rebuild. This is the right model for financial workloads where audit trails aren't optional.
+
+**eUTXO schema.** Canton contracts get created and archived, never mutated. The schema tracks the full lifecycle: create → active → archive. The ACS is a derived view.
+
+**GraphQL primary, REST + SQL secondary.** GraphQL's type system maps well to Daml templates. Subscriptions handle real-time. REST for simple lookups. SQL direct access for power users.
+
+**Schema-per-party isolation.** Each party gets its own postgres schema. Even if there's a bug in the query layer, party A's data can't leak to party B at the database level.
```
daravep, line 69: The Ledger API is already segregating the views per party and provides via its user management a per-party view on the contract state. When running PQS, you can choose to aggregate over multiple parties. In that sense, what you are proposing with this here is to undo this aggregation after the fact.

Wouldn't it just make sense to modify PQS directly such that PQS can simultaneously index multiple streams and write it into separate databases rather than undoing what you asked PQS to do for you?

proposals/canton-network-indexer.md
```diff
@@ -0,0 +1,427 @@
...
+
+### Key decisions
+
+**PQS as foundation, not a custom gRPC client.** PQS is maintained by Digital Asset. It already deals with Ledger API evolution (services got removed between 3.3 and 3.5). Building a custom ingester means tracking all of that ourselves. Not worth it unless we're past 1000 tx/s.
+
+**Event sourcing.** The raw offset stream is the source of truth. Materialized views are projections you can rebuild. This is the right model for financial workloads where audit trails aren't optional.
+
+**eUTXO schema.** Canton contracts get created and archived, never mutated. The schema tracks the full lifecycle: create → active → archive. The ACS is a derived view.
+
+**GraphQL primary, REST + SQL secondary.** GraphQL's type system maps well to Daml templates. Subscriptions handle real-time. REST for simple lookups. SQL direct access for power users.
```
daravep, line 67: Note that we used to have a GraphQL API on the Navigator backend: https://docs.daml.com/tools/navigator/index.html

proposals/canton-network-indexer.md
```diff
@@ -0,0 +1,427 @@
...
+
+PQS handles the hard parts: gRPC streaming, offset tracking, Daml-LF serialization, crash recovery. We don't want to rewrite that. Instead we extend PQS output with a post-processor that populates indexer-specific tables (contract lifecycle state, per-party activity timelines, template-specific materialized views).
+
+### Key decisions
+
+**PQS as foundation, not a custom gRPC client.** PQS is maintained by Digital Asset. It already deals with Ledger API evolution (services got removed between 3.3 and 3.5). Building a custom ingester means tracking all of that ourselves. Not worth it unless we're past 1000 tx/s.
+
+**Event sourcing.** The raw offset stream is the source of truth. Materialized views are projections you can rebuild. This is the right model for financial workloads where audit trails aren't optional.
+
+**eUTXO schema.** Canton contracts get created and archived, never mutated. The schema tracks the full lifecycle: create → active → archive. The ACS is a derived view.
```
daravep, line 65: You need to be careful here: multi-sync support is already in the code and will become GA later in the year. This means that you will have to deal with inhomogeneous topologies where contracts move in and out of view. In that sense, your concept is not multi-synchronizer ready.

proposals/canton-network-indexer.md
```diff
@@ -0,0 +1,427 @@
...
+                                                                   template mat.views
+```
+
+PQS handles the hard parts: gRPC streaming, offset tracking, Daml-LF serialization, crash recovery. We don't want to rewrite that. Instead we extend PQS output with a post-processor that populates indexer-specific tables (contract lifecycle state, per-party activity timelines, template-specific materialized views).
+
+### Key decisions
+
+**PQS as foundation, not a custom gRPC client.** PQS is maintained by Digital Asset. It already deals with Ledger API evolution (services got removed between 3.3 and 3.5). Building a custom ingester means tracking all of that ourselves. Not worth it unless we're past 1000 tx/s.
+
+**Event sourcing.** The raw offset stream is the source of truth. Materialized views are projections you can rebuild. This is the right model for financial workloads where audit trails aren't optional.
```
daravep, line 63: A minor comment on the raw offset: the offsets are validator-generated and are not global to the network. This means that aggregating the state from multiple validators cannot be based on offsets. In that sense, I don't exactly understand how you want to aggregate Scan API and Ledger API data based on offsets. 

The only concept that works is called "vector clocks" where you effectively have a map from synchronizer to record time. We do plan to add this subsequently to better support aggregation and failover between participants, but these things are not yet exposed on the API level yet.

proposals/canton-network-indexer.md
```diff
@@ -0,0 +1,427 @@
...
+CREATE MATERIALIZED VIEW mv_token_positions AS
+SELECT contract_id, payload->>'owner' AS owner,
+       (payload->>'amount')::numeric AS amount
+FROM contract_state
+WHERE template_id = 'tokenlib:Token.Holding:Holding' AND status = 'active';
+```
+
+Refreshed concurrently by the post-processor.
+
+**Week 9–10: Scan API consumer.** Separate service polling the Scan Bulk Data API. Writes to Tier 1 tables: `cc_transfers`, `minting_rounds` (TimescaleDB hypertable for time-series queries), `validator_metrics`, `governance_proposals`, `network_topology`. REST endpoints for public dashboard queries.
```
daravep, line 181: In the above you explained how you are going to separate the schema per party. Does that mean that you will be ingesting the DSO data into the "DSO party schema"?

proposals/canton-network-indexer.md
```diff
@@ -0,0 +1,427 @@
...
+
+### Milestone 3 — Aggregation + production (Weeks 11–16)
+
+**Week 11–12: Tier 3 RBAC.** Cross-party aggregation for multi-party orgs. RBAC tables: `organizations`, `roles`, `role_parties`, `user_roles`. User → resolve party set → query across authorized schemas with UNION ALL. Regulatory audit views scoped to observer visibility only.
+
+**Week 13–14: K8s + monitoring.** Helm charts for all services. Prometheus metrics: `indexer_postproc_lag_seconds`, `indexer_graphql_request_duration_seconds`, `indexer_watermark_offset`. Grafana dashboards for ingestion throughput, watermark lag, query latency, error rates.
+
+**Week 14–15: Caching + DAR schema generator.** Redis for hot contract state (TTL matched to watermark advance rate). CLI tool reads compiled DAR files, generates materialized view DDL + GraphQL types + resolvers per template. Published as npm package.
```
daravep, line 189: Note that automatically generating all kinds of schemas of DARs can quickly become a Denial Of Service on your Postgres DB server.

daravep (commented):
Thanks a lot for your proposal. I do think that there are elements where you plan to add great value:
- There is no standardized API that runs on top of the query store PQS than SQL. In that sense, you are proposing to add a consistent mapping from GraphQL / REST to SQL.
- You want to solve the "aggregating data from multiple validators".

However, I also see quite a few points where I would go with different design choices:
- You want to feed from PQS and Scan API and then harmonise the data into the same format. Why? Both of them are effectively a modified Ledger API stream. Why not just read from multiple Ledger API streams directly without the hassle? (Yes you might need to add support for this to Scan but that might be less work and give a more universal solution).
- Your design is not ready for multi-synchronizer. 
- The contract representation in the database is effectively a duplication of PQS. In that sense there is lots of data duplication and you need to re-implement the same features (e.g. pruning ...). Instead of duplicating things, you might want to consider contributing to PQS directly.
- If you don't like the PQS output schema, not that PQS was in theory architected to support different "output writers", not just the current JSONB based schema. 

Finally, I want to add a few words of caution, based on past experience:
- Building a universal index is generally hard when your index needs to scale. With 100M+ contracts and thousands of tables and parties, you will also start to push the limits of what you can properly operationalize on Postgres.
- The inhomogeneous topologies and the fact that parties can migrate between nodes make indexing non-trivial. You need to be able to deal with all these edge-cases too, otherwise you have something that only works on the happy-path. 

In that sense, my recommendation would be to reconsider your approach and discuss with the PQS maintainers on how to best implement the functionality you want to provide.

PR #228, opened 2026-04-17
Canton Foundation Hosted DevNet Nodes for Removing Friction for Developer Onboarding and Adoption.

Jatinp26:

**Proposal file:**  
Find the Proposal at https://github.com/Jatinp26/canton-dev-fund/blob/main/proposals/Foundation%20Hosted%20DevNet%20Nodes.md

## Summary

Canton Foundation will deploy and operate 3 - 4 dedicated DevNet Nodes, exposing rate-limited JSON Ledger API endpoints. Access credentials will be issued by DevRel/BD staff after a qualification review (Developer Initiative, Project legitimacy, use case fit, no bad actors, and more custom parameters). Endpoints will not appear in public documentation or on the Canton website. The Keys will be revoked after the Partner Whitelisting is done / Developer Event e.g. a Hackathon event is over, Afterwards Developers will continue on their own whitelisted IP. The goal is to eliminate the timing friction of onboarding and Building.

## Checklist

## Notes for Reviewers


proposals/Foundation Hosted DevNet Nodes.md
```diff
@@ -0,0 +1,109 @@
...
+- Keys scoped to DevNet only, non-transferable, revocable
+- Revoking API Endpoints allocated after the Partner Whitelisting is completed.
+
+**What this is not:**
+
+- Not a public sandbox, no open registration, no docs listing
+- Not a replacement for running your own validator /  Working with a NaaS, projects are expected to run their own node / Setup a NaaS for TestNet/MainNet and Get whitelisted for DevNet.
+- It's not mainnet or testnet infrastructure, only for DevNet.
+
+**Adoption and Access** (Access to these endpoints will not be publicly listed.)
```
nycnewman, line 69: Security by obscurity is never a lasting proposition.

nycnewman:
The two Github workflow ymls appear to be out of place for this proposal. Probably shoud be moved to a separate PR.

Would be good to understand issues like:
- DAR upload vetting and management for shared infrastructure
- Risks from multi-tenancy
- Options to reset to known state
- Hardening of deployment to protect against accidental as well as malicious intent
- May have missed by assumption is that APIs are exposed to Internet whilst participant is whitelisted against DevNet
- Are you expecting to suport external signing use cases?

proposals/Foundation Hosted DevNet Nodes.md
Jatinp26: Could be please expand into it?? @nycnewman

Jatinp26: if you mean these will be discovered then further down the proposal it's shared that these will be temp endpoints issued to select ones and only be valid for a limited amount of time and can be revoked if monitored a non permitted usage, like RPC points, it's on us to prevent it, but I'd recom to overlook "Security by obscurity" as that is case with each NaaS provider we will work with as well as the aim is the solve the onboarding fiction identified clearly by the DevRel team over last quarter.

Jatinp26:
> The two Github workflow ymls appear to be out of place for this proposal. Probably shoud be moved to a separate PR.
> 
> Would be good to understand issues like:
> 
> * DAR upload vetting and management for shared infrastructure
> * Risks from multi-tenancy
> * Options to reset to known state
> * Hardening of deployment to protect against accidental as well as malicious intent
> * May have missed by assumption is that APIs are exposed to Internet whilst participant is whitelisted against DevNet
> * Are you expecting to suport external signing use cases?

- removed the other yaml files, just testing new flow foundation is building for dev fund and got added to check against the main branch too. **Removed them.**

- For security side, we are aiming to work with a trusted NaaS for this to tackle current above given as well as other potential security concerns as well (given in the proposal) that said, still adding some info how this generally gets tackled:
     - DAR uploads on a shared participant are gated by the access model itself keys are issued only after qualification review by DevRel/BD, so anonymous or unvetted actors can't reach the Ledger API in the first place. For the NaaS-operated nodes, DAR vetting policy will follow the operator's standard multi-tenant controls, which we'll document as part of the M2 access control guidelines. The short lived key model (14-day auto-revocation) also limits blast radius and a bad actor's window is bounded by key validity, not by the lifetime of the node.
     - The NaaS provider we'll working with will operate multi-tenant Canton infrastructure commercially today as this isn't a new problem we're solving from scratch, it's a configuration we're applying to a known pattern. Infrastructure-level isolation is handled at the API layer via key rate limits on requests/minute and transaction volume caps which is a M2 deliverable. Also, Afik Canton's privacy model handles ledger-level isolation natively that is Party A cannot observe Party B's contracts by protocol design, not by configuration.
     - DevNet itself resets periodically per network's schedule which automatically returns nodes to a clean state. For party-level resets between events (e.g. between hackathon cohorts) the key revocation model handles these old keys are revoked, new keys issued to the next cohort and the onboarding state is clean from the developer's perspective. 
     - Hardening against accidental and malicious intent: already specified in proposal with access qualification before key issuance by FOUNDATION THEMSELVES, per-key rate limiting, NaaS level hardening on the node itself, and only the JSON Ledger API endpoints scoped to the issued party credentials.
     - Developers get a clean API endpoint to build against without needing their own whitelisted IP which is precisely the friction this proposal eliminates. The API gateway sits between the internet and the participant, handling auth, rate limiting, and logging before anything touches the Canton node. this is as secure as anything already on canton validator node side.
     - for External signing, as already given, Not in scope for this proposal as use cases here are early-stage development and developer activation without friction of whitelisting e.g hackathons neither of which requires custody grade external signing workflows. Thanks
     
Overall, I believe security vise it's already covering all above pointers. Thanks!

shaul-da:
@Jatinp26 have we considered an alternative where we use a grant to fund some of the existing NaaS operators to do this? Seems like a duplication of efforts, and puts Foundation personnel on the critical path to keep infrastructure up and up-to-date

Jatinp26:
Hey @shaul-da thanks for the suggestion, it’s something we did considered.

- A concern with that approach is keeping the Foundation neutrality. Routing this through a specific NaaS provider risks creating a perception of a preferred vendor, which we avoid across the ecosystem as a Foundation.

- Plus, this is intended to be a lightweight, developer-focused resource for early-stage teams and dev advocacy initiatives like hackathons. Introducing a third-party layer tends to increase both cost and coordination overhead for foundation.

The grant structure doesn’t prevent collaboration with NaaS providers where it makes sense for sure, it just keeps the Foundation from being in a position of having to select one by allowing the grant to execute this.

proposals/Foundation Hosted DevNet Nodes.md
```diff
@@ -0,0 +1,132 @@
...
+**Author:** Canton Foundation  
+**Status:** Draft  
+**Created:** 2026-04-15  
+**Label:** node-deployment-operations
+
+**Champion:** Canton Foundation
+
+## Abstract
+
+Canton Foundation will deploy and operate 3–4 dedicated DevNet validator nodes, exposing rate-limited JSON Ledger API endpoints. Access credentials will be issued by DevRel/BD staff after a qualification review (project legitimacy, use case fit, no bad actors). Endpoints will not appear in public documentation or on the Canton website. JWT tokens will be revoked after Partner Whitelisting is done / Hackathon event is over i.e Foundation will issue time-limited JWT tokens which automatically expire after 14 days. The qualification bar is intentionally low, the goal is to eliminate the timing friction of onboarding and Building.
```
leonidr-c7, line 12: Why choose 3-4 validator nodes? I think the key concept that you want to support is concurrent developers interacting with Canton Network, so you if you want to support 1000 independent developers, you want to support 1000 Party’s allocated on some nodes in the network? Presumably there is an optimal, from cost and management perspective, way to achieve that.

The rest of your proposal (correctly imo) restricts how users will interact with Canton Network (Daml application development, not infrastructure, Canton internal specific), so I think `Party` access would be a better metric.

proposals/Foundation Hosted DevNet Nodes.md
```diff
@@ -0,0 +1,132 @@
...
+
+**Infrastructure:**
+
+- 3 – 4 validator nodes deployed on Canton DevNet, meeting standard requirements.
+- Nodes operated by a trusted NaaS provider under Canton Foundation direction. Canton Foundation retains admin rights over the participant nodes. The NaaS provider manages infrastructure operations (compute, uptime, upgrades) and executes user/party management actions at Foundation direction only.
+
+**Access control:**
+
+- JWT token issuance managed by DevRel/BD team via an internal tracking sheet.
+- Per-token rate limits enforced at a access controll layer (requests/minute, transaction volume caps, and more!)
```
leonidr-c7, line 37: How will you implement this, and where will the logic run?

proposals/Foundation Hosted DevNet Nodes.md
```diff
@@ -0,0 +1,132 @@
...
+
+### 2. Implementation Mechanics
+
+**Infrastructure:**
+
+- 3 – 4 validator nodes deployed on Canton DevNet, meeting standard requirements.
+- Nodes operated by a trusted NaaS provider under Canton Foundation direction. Canton Foundation retains admin rights over the participant nodes. The NaaS provider manages infrastructure operations (compute, uptime, upgrades) and executes user/party management actions at Foundation direction only.
+
+**Access control:**
+
```
leonidr-c7, line 35: (Apologies, this comment belongs on line 36, but you know github and UI’s 🤦 ).

> JWT token issuance managed by DevRel/BD team via an internal tracking sheet

Will you create a program to create and track these? I wonder if this is easy for a Keycloak 🥷 ?

leonidr-c7 (commented):
This proposal addresses a legitimate and useful need, but some of the details could be better specified.

hythloda:
Come back after implemented with budget request

PR #297, opened 2026-05-04
Proposal: Canton Failure-Classification Engine (Canton Transaction Debugger Earlier)

infrasingularity:
### Development Fund Proposal Submission

**Proposal file:** `/proposals/canton-failure-classification-engine.md`

**Summary**

A Canton-native failure-classification engine that turns failed or rejected Ledger API submissions into deterministic, categorized root-cause diagnoses with ranked next checks — backed by an open, community-extensible knowledge base that the ecosystem can grow, and delivered as a **DPM component** so it installs through the toolchain Daml teams already use.

Existing Canton tooling operates on the success path: it inspects committed transactions by update-id and streams ledger events, all of which assume a transaction exists on the ledger to examine. When a submission is rejected, no update-id is created and nothing commits, and the developer reconstructs root cause by hand from the rejection payload and error code. That is where Canton is hardest to reason about, because its most painful failures — authorization, sub-transaction privacy/visibility, package/vetting drift, contention — are often the ones that never commit.

This build starts from the failed submission rather than the successful one. It parses Canton's structured error codes and rejection payloads and classifies each failure into a named family with ranked next checks at a measured precision. Where the available signals cannot distinguish the cause, it returns `insufficient signal` rather than guessing — the tool is bounded honestly to what the Ledger API exposes. Anyone can extend the knowledge base by contributing a new diagnosis (a declarative rule plus test examples) through a reviewed pull request, with no engine changes required: decoder packs are published as **data-only DPM components that export under a shared key with `conflict-strategy: extend`**, so `dpm` discovers and concatenates canonical, experimental, and organization-private packs while the engine arbitrates between them deterministically. It builds on the existing toolchain — Ledger API rejection payloads and error codes, plus `dpm`'s published component and export mechanisms — with no compiler dependency, no fork of any component, and **no bespoke plugin registry, loader, or version-resolution machinery of its own**. It complements the DPM diagnostic component family — DPM Trace (#327), DPM Debug (#494), and the Transaction Profiler (#379) — all of which operate on transactions that exist.

**Total funding request:** **450,000 CC** across **3 milestones** (~10 weeks).

**Checklist**
- Proposal file added under `/proposals/`
- Milestones and funding amounts defined
- Acceptance criteria included
- Alignment with Canton priorities described

**Notes for Reviewers**

**Label:** `daml-tooling`

This is a single-objective build: a failure-classification engine plus the open, community-extensible knowledge base behind it. Community extensibility is the core of the design, per SIG feedback — contributions are declarative (no contributed code runs in the engine, and packs are data-only components), fixture-gated with an automated no-regression check, deterministically resolved, and namespaced (canonical / experimental / org-private) with governed acceptance via the SIG. Two guarantees are enforced mechanically rather than asserted: **version-gating is verified in CI across a matrix of at least 3 SDK versions installed through `dpm`** — every decoder must classify correctly inside its declared range and abstain outside it, so knowledge-base rot surfaces as a build failure — and **namespace precedence is enforced against a signed canonical index**, so an untrusted pack cannot outrank a reviewed decoder through `extend` concatenation.

Delineation against adjacent work is deliberate: PRs #327, #494, and #379 all operate on committed transactions (inspection, visual debugging, execution cost); this engine addresses submissions that never committed, where there is no transaction to inspect. The only overlap is decoding a rejection against DAR metadata. A stable **integration contract** — library interface plus versioned JSON diagnosis schema, with a reference adapter — is published so those components can call a measured classifier instead of each reimplementing failure heuristics.

Funding is split 50% on delivery and 50% on independently verified outcomes (measured diagnostic precision/coverage, time-to-diagnosis, and validation on real production failures), so half the total — **225,000 CC** — is gated on measured ecosystem value rather than artifact delivery, with the outcome share rising across milestones.

The team (InfraSingularity) ships Daml-based projects and operates production Canton validator infrastructure as an institutional NaaS provider; the diagnosis knowledge base is seeded from real production failures. Relevant SIGs: Daml Language & Developer Tooling, Canton APIs, dApp Integration.

**Unique solutions** (technical detail in the proposal file):
- Failure-first diagnosis: starts from the rejected submission, not the committed transaction
- Deterministic, categorized diagnosis at a measured precision, with honest `insufficient signal` when the signal is inadequate
- Community-extensible knowledge base riding `dpm`'s native `exports` / `extend` mechanism — data-only packs, no contributed code executes, private packs via standard registry credentials, no fork required
- Loophole-closed governance: fixture-gated no-regression CI, deterministic conflict resolution, signed-canonical-index namespace precedence, and version-gating **verified** across a ≥3-version SDK matrix
- Non-invasive integration: Ledger API and `dpm`'s published component mechanism only — no compiler dependency, no fork, no parallel registry
- Published integration contract so the DPM diagnostic components (#327, #494, #379) consume classification rather than rebuild it

Full technical specifications and per-milestone deliverables/acceptance criteria are in the proposal file.

**Champion:** @v9n


proposals/canton-transaction-debugger.md
```diff
@@ -0,0 +1,283 @@
...
+
+Reduce time-to-root-cause for failed Canton transactions in privacy- and authorization-heavy scenarios from hours to minutes, by producing a **deterministic decoded diagnosis** and a **sanitized, shareable debug bundle** for every failed submission.
+
+Single objective: build the debugger and its bundle artifact. Replay is a small, bounded inclusion within Milestone 3 for a clearly listed category — not a competing objective, and not gating MVP value.
+
+### 2. Implementation Mechanics
+
+Four components, all sitting on top of the existing Canton Ledger API surface:
+
+- **Trace Collector** — ingests rejection payloads and structured client/gateway markers keyed by a client-supplied correlation ID; assembles a normalized Trace Artifact; applies redaction and TTL.
```
v9n: Just curious a concrete idea/plan to do this? By both ingest the ledger through v2/updates and enrich the data with the log in debug mode?

proposals/canton-transaction-debugger.md
```diff
@@ -0,0 +1,283 @@
...
+- **Decoding Engine** — rules-based and deterministic. A registry of decoders maps Canton/Daml rejection families (auth/scope, act-as/read-as, disclosure/visibility, malformed command, missing choice args, package vetting drift, environment drift) to a primary diagnosis sentence + 2–5 ranked next checks. Unknown cases return `insufficient signal` — the engine never fabricates a category. No LLM dependency for correctness.
+- **Debugger UI / API** — opens a Debug Session by correlation ID and shows the decoded diagnosis, a sanitized payload viewer, a stage timeline (only **observed** stages; missing stages are explicitly labeled "not observable"), and a one-click bundle export. Tenant-scoped permalinks with access checks.
+- **Replay CLI (scoped follow-on)** — consumes a Debug Bundle and replays a small set of supported categories on localnet (e.g. malformed command, scope/header issues, package-id mismatch). Unsupported categories are deterministically refused with a clear reason.
```
v9n: How do this replay work on localnet given many thing are different?

Example, let say a failed transfer of CIP-56 because the recipient has not vetted the same dar?

- synchronizer id
- different parties

proposals/canton-transaction-debugger.md
infrasingularity: The plan is to make the command completion and error path the primary source of truth for failed submissions, rather than relying on /v2/updates.
When a command is rejected, the most useful information comes directly from the submission and completion flow:
Submission metadata such as command_id, submission_id, correlation IDs, parties, template information, synchronizer preferences, and other sanitized command details.
Completion results including the rejection status, error payload, Canton error category/code, retryability information, trace context, synchronizer time, and related identifiers.
Environment information such as package IDs, package-vetting state, connected synchronizers, client/participant versions, and relevant topology metadata.
Optional enrichment from structured logs or distributed tracing when those signals are available.
/v2/updates remains useful as a secondary source for additional context, such as nearby successful transactions, topology changes, or committed ledger activity. However, for a failed command, it typically does not contain the primary failure signal because that signal exists in the completion/error path.
This also aligns with the proposal's intended signal contract. The debugger is designed around Ledger API completion statuses, rejection payloads, correlation IDs, and environment metadata. Gateway markers and participant logs are treated as optional enhancements, not requirements.
We should also avoid depending on Canton DEBUG-level logging. If detailed logs are available, they can enrich the diagnosis. If not, the UI should simply display the stages that were observed and clearly mark unobservable stages as "not observable" rather than attempting to infer participant-internal behavior.

proposals/canton-transaction-debugger.md
infrasingularity: The goal of localnet replay is not to recreate production state byte-for-byte. Instead, replay is intended to reproduce the same class of failure whenever the debug bundle contains enough safe metadata to do so.

The replay process would work by mapping production-specific details into a controlled local environment:

1.Production parties are replaced with anonymized local parties that preserve the relevant roles and permissions.
2.The original synchronizer ID is retained as metadata, but mapped to a local synchronizer.
3.Package IDs, package names, versions, hashes, and vetting information are used to recreate an equivalent local package/vetting setup.
4.Sensitive contract data and private payloads are redacted or replaced with minimal fixture data.

Replay is therefore not expected to reproduce the exact production transaction ID, synchronizer ID, party IDs, contract state, transaction hash, or private payloads. The success criterion is reproducing the same diagnosis category and remediation guidance.
For the CIP-56 package-vetting example, this could be a supported replay category if package-vetting drift is explicitly included in the replay-supported set and the debug bundle contains the required package and vetting metadata. In that case, the replay environment can create a minimal sender/recipient setup, upload and vet the required DAR on the sender’s participant/validator, intentionally leave the recipient’s participant/validator without the required vetted package, and execute a minimal transfer flow that is expected to fail for the same package-vetting reason.
If a failure depends on production-only topology, unavailable DARs, private contracts, or information intentionally redacted from the debug bundle, replay should fail deterministically with a clear explanation instead of pretending to reproduce the issue. For example:
Replay unsupported: insufficient safe metadata to reproduce recipient-side package-vetting state.
In those cases, the debugger still provides diagnostic analysis based on the captured failure signals, but does not attempt localnet replay.

Jatinp26:
@infrasingularity 
- Could you please share why this needs to be a separate setup and not a dpm component? It'd be better to have it integrated with current smart contract framework.
- 700,000 CC (~37% of total) for public launch + adoption campaign + 5 pilot organizations + testimonials + impact report. This is the largest single milestone and it's predominantly DevRel and marketing work, The dev fund is majorly for funding the development and adoption but there seems to be a major budget ask for marketing alone. 
- Some institutional ref there like Goldman, how'd the team committing to work with them, if it's conditional then M3's overall approach should be boiled down to more fixed adoption metrics.

infrasingularity:
> @infrasingularity
> 
> * Could you please share why this needs to be a separate setup and not a dpm component? It'd be better to have it integrated with current smart contract framework.
> * 700,000 CC (~37% of total) for public launch + adoption campaign + 5 pilot organizations + testimonials + impact report. This is the largest single milestone and it's predominantly DevRel and marketing work, The dev fund is majorly for funding the development and adoption but there seems to be a major budget ask for marketing alone.
> * Some institutional ref there like Goldman, how'd the team committing to work with them, if it's conditional then M3's overall approach should be boiled down to more fixed adoption metrics.

1. Three structural reasons this can't live there:
- No live session model: the debugger opens a session from a correlation ID in a real completion stream; dpm has no concept of a running network session.
- Cross-org sharing is the core feature: two counterparties with different visibility into the same failure need to exchange redacted bundles independent of either side's local toolchain. Inside dpm, both sides would need matching installs just to trade a bundle, which defeats the open/versioned schema this is designed around.
- It's not ours to place there: dpm is DA's own SDK CLI on DA's release train. For contrast — Certora's Daml Package Analyzer (this same fund, announced last month) *is* built into dpm, because it's static analysis of compiled .dar files: a build-time question. Ours is a runtime one — what actually happened on the network after a submission already failed there.
2.
M4 adoption phase — even before this revision — was a smaller share of total budget (37%) than the closest comparable tooling grant in this repo, Certora's Daml Package Analyzer, which allocates 83% of its budget to adoption milestones. That said, revisiting the effort and timeline for M4 specifically, we're bringing it down further to 650,000 CC (35% of a new 1,850,000 CC total). Proposal file updated to match.
3.
To confirm: M4's acceptance criteria don't name or require Goldman specifically — payment depends on the five-independent-organization bar, the 3-4 testimonials, and the published Impact Report, regardless of which teams participate. The Goldman reference is descriptive context on how the pilot pipeline is forming, not a funding condition. We have tighten the Note's wording so it can't be read as a commitment.

Jatinp26:
for:
> It's not ours to place there: dpm is DA's own SDK CLI on DA's release train. For contrast — Certora's Daml Package Analyzer (this same fund, announced last month) is built into dpm, because it's static analysis of compiled .dar files: a build-time question. Ours is a runtime one — what actually happened on the network after a submission already failed there.

- Dpm components are already open for community to build and its announced already on the forum, https://forum.canton.network/t/dpm-components-extend-the-canton-developer-stack/8822

> No live session model: the debugger opens a session from a correlation ID in a real completion stream; dpm has no concept of a running network session.

- Can spin it up in a localview UI doing the same I believe like other txn debugger proposals coming as dpm component.

> M4 adoption phase — even before this revision — was a smaller share of total budget (37%) than the closest comparable tooling grant in this repo, Certora's Daml Package Analyzer, which allocates 83% of its budget to adoption milestones. That said, revisiting the effort and timeline for M4 specifically, we're bringing it down further to 650,000 CC (35% of a new 1,850,000 CC total). Proposal file updated to match.

- Understandable, since it's about adoption metric based payment, what are the criteria and success metric via which committee can verify if it's succussed in adoption? it needs more input and data based adoption metric to be measured.

infrasingularity:
> for:
> 
> > It's not ours to place there: dpm is DA's own SDK CLI on DA's release train. For contrast — Certora's Daml Package Analyzer (this same fund, announced last month) is built into dpm, because it's static analysis of compiled .dar files: a build-time question. Ours is a runtime one — what actually happened on the network after a submission already failed there.
> 
> * Dpm components are already open for community to build and its announced already on the forum, https://forum.canton.network/t/dpm-components-extend-the-canton-developer-stack/8822
> 
> > No live session model: the debugger opens a session from a correlation ID in a real completion stream; dpm has no concept of a running network session.
> 
> * Can spin it up in a localview UI doing the same I believe like other txn debugger proposals coming as dpm component.
> 
> > M4 adoption phase — even before this revision — was a smaller share of total budget (37%) than the closest comparable tooling grant in this repo, Certora's Daml Package Analyzer, which allocates 83% of its budget to adoption milestones. That said, revisiting the effort and timeline for M4 specifically, we're bringing it down further to 650,000 CC (35% of a new 1,850,000 CC total). Proposal file updated to match.
> 
> * Understandable, since it's about adoption metric based payment, what are the criteria and success metric via which committee can verify if it's succussed in adoption? it needs more input and data based adoption metric to be measured.



Yes, the DPM components extension model announced on the forum is great for build-time/static tooling. Our debugger, however, is a runtime tool that requires live correlation-ID sessions from real completions, per-party privacy redaction for cross-org shareable bundles, and observed-only trace reconstruction from production logs. These go beyond current DPM capabilities. The open bundle schema enables broad ecosystem adoption (CI, ticketing, IDEs) while keeping future DPM/LocalView integration straightforward.

The evaluation completion based on ecosystem value, not artifact delivery alone:

* **Adoption:** at least **5 independent organizations/teams** integrate the debugger and produce real (non-fixture) bundles within the grant period, of which no more than **2** may be existing InfraSingularity NaaS clients.
* **Diagnostic effectiveness (Milestone 3 gate):** for the 6–8 covered failure categories, decoders return a correct primary diagnosis on a published fixture corpus with **≥ 90% precision**, with explicit "unknown" outcomes for everything else (no fabricated categories).
* **Live diagnostic accuracy (Milestone 4 gate):** of real bundles generated during the adoption window, **≥ 80%** receive an "accurate" rating from the reporting user via a one-click in-UI control, aggregated automatically rather than inferred from testimonials.
* **Time-to-root-cause:** tool-timestamped from session-open to a "mark resolved" action, measured against each participating team's own pre-registered baseline estimate — **median drops below 30 minutes**, or **≥ 70% reduction** vs. baseline, for covered categories.
* **Support load:** participating teams answer two structured, numeric survey questions (escalations per month, before vs. during pilot) rather than a general impression.
* **Privacy posture:** redaction policy review signed off; cross-tenant access tests pass; TTL enforcement verified.

Jatinp26:
@infrasingularity 
following up on the dpm component question with more specifics:

I wanna push back on this tooling as a fragmented dpm tool since Transaction Debugger Technically Possible as a dpm Component:

- The argument that "dpm has no concept of a running network session" isn't accurate given the current dpm toolchain i.e. dpm sandbox runs a full live Canton node and dpm daml-shell opens a live interactive session against PQS also dpm pqs starts a live service. These are all runtime tools, not build-time tools, and they all ship as dpm components.

- The community dpm component model (announced on the forum) exists specifically so third-party teams can publish components and share with community for public good in a right way, So the "not ours to place there" argument also doesn't apply as you'd publish canton-debugger as a community component, same as other teams so far e.g daml package manager by certora.

- The cross-org bundle sharing concern is real but it's not an argument against being a dpm component since the bundle schema can be an open file format regardless of how the tool is invoked AFIK dpm canton-debugger export produces a .zip that anyone can share.

**The concrete question I'd like answered before vote is** has the team evaluated implementing the trace collector + decoding engine + bundle export as a dpm component, with an optional hosted web frontend on top?

If there's a genuine technical reason this doesn't work within the dpm component model after evaluating it, I'd like to hear it specifically so we can vote without any blocker on why fragment a dpm tooling from dpm.

infrasingularity:
> @infrasingularity following up on the dpm component question with more specifics:
> 
> I wanna push back on this tooling as a fragmented dpm tool since Transaction Debugger Technically Possible as a dpm Component:
> 
> * The argument that "dpm has no concept of a running network session" isn't accurate given the current dpm toolchain i.e. dpm sandbox runs a full live Canton node and dpm daml-shell opens a live interactive session against PQS also dpm pqs starts a live service. These are all runtime tools, not build-time tools, and they all ship as dpm components.
> * The community dpm component model (announced on the forum) exists specifically so third-party teams can publish components and share with community for public good in a right way, So the "not ours to place there" argument also doesn't apply as you'd publish canton-debugger as a community component, same as other teams so far e.g daml package manager by certora.
> * The cross-org bundle sharing concern is real but it's not an argument against being a dpm component since the bundle schema can be an open file format regardless of how the tool is invoked AFIK dpm canton-debugger export produces a .zip that anyone can share.
> 
> **The concrete question I'd like answered before vote is** has the team evaluated implementing the trace collector + decoding engine + bundle export as a dpm component, with an optional hosted web frontend on top?
> 
> If there's a genuine technical reason this doesn't work within the dpm component model after evaluating it, I'd like to hear it specifically so we can vote without any blocker on why fragment a dpm tooling from dpm.


We have re-evaluated and will adapt the proposal: the core functionality (Trace Collector + Decoding Engine + Bundle Export) will be delivered as a DPM community component (dpm canton-debugger), following the model announced on the forum and used by Certora. This allows seamless CLI use in local/dev workflows (including sandbox, PQS, etc.).
The full live debug sessions, web UI, and production deployment model (especially privacy redaction for cross-org bundles in institutional/NaaS setups) will remain available as an optional hosted service on top. The bundle schema stays fully open.
We will update the proposal with this DPM-first approach, including M3 deliverables for component packaging. This resolves the fragmentation concern while preserving the tool’s value for production use cases.

proposals/canton-transaction-debugger.md
```diff
@@ -0,0 +1,288 @@
...
+
+**2. Trace-id log correlation across nodes (best-effort, observed-only).**
+Canton propagates OpenTelemetry trace context (trace-id / span-id) through the request lifecycle and emits it in structured (JSON / logback) logs on participant and synchronizer (sequencer, mediator) nodes. **Where a deployment exposes these logs — an optional, best-effort signal, never assumed** — the tool filters by the trace id taken from the completion, orders the spans causally, and reconstructs as much of the lifecycle (submission → confirmation request → confirmation responses → mediator verdict → completion) as is **actually observable**; any stage absent from the available signal is labelled "not observable" rather than inferred. This keeps the feature consistent with the Signal Contract in §2: node logs are never a hard dependency. The correlation logic is built against the log artifacts production Canton validators actually emit, drawn from InfraSingularity's live operation, not from assumed formats.
+
+**3. Privacy-aware, shareable debug bundles.**
```
hrischuk-da: The completion stream is only on the submitting / executing validator.   Other failures can occur in the preparation stage.   I'm struggling to understand what value the debug bundle provides.  Can you explain 3 use cases where it would help?

proposals/canton-transaction-debugger.md
```diff
@@ -0,0 +1,288 @@
...
+
+**4. Prepared-vs-failed diff for interactive submissions** *(additive future extension — not a funded milestone; out of scope for M1–M6).*
```
hrischuk-da: Is this part of the project because I interpret it as out of scope.   If so, why include it?

infrasingularity: We intentionally kept the schema forward-compatible so the solution can evolve into a more complete, production-ready, and market-fit diagnostic product without expanding the scope of the current proposal.

proposals/canton-transaction-debugger.md
infrasingularity: Thank you for the feedback.
First, the completion stream is scoped to the participant that submitted or executed the command. The proposal does not assume that other parties can access or observe another participant's completion stream.
Second, preparation-stage failures are returned as synchronous gRPC errors from the prepare call rather than as completion events. Our Decoding Engine already accounts for this by ingesting synchronous preparation errors (Unique Solutions 1, input (a)), ensuring these failures are still decoded and included in the diagnostic bundle. The only capability intentionally out of scope is generating a prepared-versus-executed comparison, as described in 4.
We'll also revise 3 to better emphasize that the primary value of the diagnostic bundle is on the submitter side, with cross-party sharing being an optional, one-way workflow rather than a symmetric exchange. Some representative use cases are:
* Intra-organization escalation: An application team encounters an authorization, package, or topology-related rejection and exports a single diagnostic bundle containing the decoded failure, sanitized request context, and prioritized troubleshooting guidance. This bundle can be attached directly to an internal support ticket, allowing infrastructure or platform engineers to diagnose the issue without requesting additional logs or reproducing the failure. This aligns with the operational cost reduction highlighted in the Motivation section.
* Submitter-to-counterparty troubleshooting: A multi-party transaction may fail due to configuration or topology issues on another participant (for example, missing disclosures or controller authority). The submitter can share a redacted diagnostic bundle containing only the error category, correlation identifier, and high-level template or choice information, without exposing contract payloads or participant-private data. The receiving party can use this information to validate its own configuration while correlating the incident through the shared identifier.
* Cross-environment diagnostics: When an integration succeeds in one environment but fails in another, the submitter can export diagnostic bundles from each environment and compare metadata such as package versions, topology configuration, participant software versions, and client configuration. This enables rapid identification of environment-specific drift without requiring participant logs or cross-party visibility.
To avoid implying bidirectional access or information exchange, we'll update the wording in 3 to focus on the submitter-centric workflow and present cross-party sharing as an optional, controlled, and one-directional mechanism.

shaul-da:
It seems this proposal covers a similar scope to https://github.com/canton-foundation/canton-dev-fund/pull/327
I've consulted with our team, and they decided to champion the other proposal as they see it as technically stronger; I'll likely vote for it rather than this one.

@infrasingularity and @v9n, I recommend you take a look at that proposal and let us know if you think there's still value in your proposal beyond it. Or if it makes sense to carve out specific parts of your proposal as additions to the DPM Trace source-aware debugger.

infrasingularity:
> It seems this proposal covers a similar scope to #327 I've consulted with our team, and they decided to champion the other proposal as they see it as technically stronger; I'll likely vote for it rather than this one.
> 
> @infrasingularity and @v9n, I recommend you take a look at that proposal and let us know if you think there's still value in your proposal beyond it. Or if it makes sense to carve out specific parts of your proposal as additions to the DPM Trace source-aware debugger.

Having gone through its current scope, the two overlap on one layer and diverge on the rest.

Both decode a Canton transaction against DAR metadata and can export a bundle. Beyond that they do different jobs. DPM Trace is an interactive inspector — it navigates an update's tree, diffs prepared / committed / failed states, and links to Daml source where metadata allows. That's real value and we don't want to duplicate it.

What #297 adds sits in the areas DPM Trace doesn't cover, and all are substance of this proposal rather than nice-to-haves:

1. A failure-classification engine. A decoder registry maps a failed submission to a known rejection family — auth/scope, act-as/read-as, disclosure/visibility, malformed command, package/vetting drift — and returns a categorized root cause with ranked next checks, against a stated precision target. This turns "here is the decoded transaction" into "here is which of N known failure modes this is and what to check first." DPM Trace renders and diffs; it does not classify, and it commits to no diagnosis-quality metric.

2. A governed, shareable escalation artifact. The debug bundle is redacted, versioned, tenant-partitioned, and TTL-governed, with access checks and cross-tenant isolation — so it can be attached to a ticket or handed across a support boundary without exposing counterparty data. DPM Trace can export a bundle, but there is no redaction, retention, or tenant-isolation model around it. For institutional Canton operators that governance layer is not optional, which is why we'd rather build it properly than drop it.

The ecosystem still has no answer for pre-execution failures, no cross-org sharing, and no measured diagnostic accuracy — all three are named gaps in the Feb 2026 DevEx survey.

Given DPM Trace is DPM-native and already owns the inspection and diff UX, we think the right structure is a split rather than a merge or a withdrawal. We'd re-scope #297 so that the decoder registry and the redaction / bundle-governance model (components for DPM Trace and other tooling can consume) , and the classification engine and the privacy / tenant layer are declared as the funded deliverable here. That removes the overlap, gives DPM Trace the failure-diagnosis and governed-sharing capabilities it currently lacks, and preserves the parts of this proposal that have no counterpart in #327. 
Glad to define the exact component boundary with the Walnut team.

We have updated the proposal file accordingly and have also reduced our CC ask significantly seeing there is change in the milestones we need to produce, rest the format remains the same follows- all adaptation guidelines and highlights the " Need of our proposal" in the ecosystem without contradicting with any others in progress.

@Jatinp26 @shaul-da @hrischuk-da

hrischuk-da:
@infrasingularity I agree that there are two other components to this proposal and I suggest you split them out as separate grant proposals.  I do have some additional thoughts:

- A failure-classification engine.   I would expect that it is extensible in a way that the community can build on it.   This would mean providing mechanisms to add to the knowledge base and diagnosis.
- A governed, shareable escalation artifact -- the debug bundle.  As I mentioned in my comment, it is not clear to me the additional value this brings as a sharable item across validators for debugging since the client doing the submitting should see (I think) all the information that the other parties can.  In fact, the client is the only one to see errors at the preparation stage.  

I think the failure classification engine could have high value.

infrasingularity:
> @infrasingularity I agree that there are two other components to this proposal and I suggest you split them out as separate grant proposals. I do have some additional thoughts:
> 
> * A failure-classification engine.   I would expect that it is extensible in a way that the community can build on it.   This would mean providing mechanisms to add to the knowledge base and diagnosis.
> * A governed, shareable escalation artifact -- the debug bundle.  As I mentioned in my comment, it is not clear to me the additional value this brings as a sharable item across validators for debugging since the client doing the submitting should see (I think) all the information that the other parties can.  In fact, the client is the only one to see errors at the preparation stage.
> 
> I think the failure classification engine could have high value.

Thanks for the steer — we've reworked the proposal along these lines, and the file is updated on this PR, it's now scoped to the single component you flagged as high-value — the failure-classification engine — and the debug bundle is removed entirely.

We've also built in the community extensibility you asked for: anyone can add a new diagnosis by opening a pull request (a simple rule plus test examples), which is automatically checked and reviewed through the Daml tooling SIG before it joins the shared knowledge base — no need to touch the engine itself.

The scope now runs across three milestones:

- M1 — the open building blocks: the failure taxonomy, the format teams use to contribute a diagnosis, and a public test corpus (seeded from real production failures) to measure accuracy against.
- M2 — the engine itself, plus an initial set of 6–8 diagnoses covering the common Canton failure categories, delivered as a CLI and API.
- M3 — the community contribution pipeline, docs and a starter guide, and validation on our own production failures.

The result is a single, smaller, focused deliverable. Happy to walk the SIG through the contribution format, and to line the open pieces up with the DPM Trace team so there's no overlap.

@Jatinp26  @hrischuk-da @shaul-da

infrasingularity:
> > @infrasingularity I agree that there are two other components to this proposal and I suggest you split them out as separate grant proposals. I do have some additional thoughts:
> > 
> > * A failure-classification engine.   I would expect that it is extensible in a way that the community can build on it.   This would mean providing mechanisms to add to the knowledge base and diagnosis.
> > * A governed, shareable escalation artifact -- the debug bundle.  As I mentioned in my comment, it is not clear to me the additional value this brings as a sharable item across validators for debugging since the client doing the submitting should see (I think) all the information that the other parties can.  In fact, the client is the only one to see errors at the preparation stage.
> > 
> > I think the failure classification engine could have high value.
> 
> Thanks for the steer — we've reworked the proposal along these lines, and the file is updated on this PR, it's now scoped to the single component you flagged as high-value — the failure-classification engine — and the debug bundle is removed entirely.
> 
> We've also built in the community extensibility you asked for: anyone can add a new diagnosis by opening a pull request (a simple rule plus test examples), which is automatically checked and reviewed through the Daml tooling SIG before it joins the shared knowledge base — no need to touch the engine itself.
> 
> The scope now runs across three milestones:
> 
> * M1 — the open building blocks: the failure taxonomy, the format teams use to contribute a diagnosis, and a public test corpus (seeded from real production failures) to measure accuracy against.
> * M2 — the engine itself, plus an initial set of 6–8 diagnoses covering the common Canton failure categories, delivered as a CLI and API.
> * M3 — the community contribution pipeline, docs and a starter guide, and validation on our own production failures.
> 
> The result is a single, smaller, focused deliverable. Happy to walk the SIG through the contribution format, and to line the open pieces up with the DPM Trace team so there's no overlap.
> 
> @Jatinp26 @hrischuk-da @shaul-da



Hey @Jatinp26 @hrischuk-da @shaul-da , hope you're all doing well!
Just wanted to check in to see if you've had a chance to review the latest changes. Please let us know if there's anything else you'd like us to address before Wednesday's vote.
This proposal has been in motion for several months now, and we've worked hard to incorporate all the feedback and suggestions we've received along the way. We're really excited about where it's landed and are eager to get started as soon as possible.
Thanks again for all your time, feedback, and support throughout the process—we really appreciate it!


Hey @Jatinp26 @hrischuk-da @shaul-da , we've revised our proposal, incorporated all the changes based on the last round of feedback, and also reduced the CC amount.
We'd really appreciate it if you could take another look and share your feedback. If you feel there's still anything we've missed or any further improvements we should make, please let us know—we're happy to address any additional suggestions. Thanks!

proposals/canton-failure-classification-engine.md
```diff
@@ -0,0 +1,238 @@
...
+**Created:** 2026-05-04
+**Updated:** 2026-07-30
+**Label:** `daml-tooling`
+**Champion:** TBD
+
+---
+
+## Abstract
+
+A Canton-native, **community-extensible failure-classification engine** that turns failed or rejected Ledger API submissions into a deterministic, categorized root cause with ranked next checks — and lets any team contribute new decoders and diagnoses to a shared, versioned **knowledge base** without engine access. Seeded from real production Canton failures and grown by the ecosystem, it aims to move time-to-root-cause on covered failures from hours to minutes.
```
daravep, line 14: Can you elaborate how this compares to the existing complete error code inventory with explanation and resolution? https://docs.canton.network/global-synchronizer/reference/error-codes#error-code-reference

The purpose of the structured error codes is to ensure that everything is documented at the source in order to guarantee consistency. 

Unless I misunderstand I don't really see any reason to duplicate this.

proposals/canton-failure-classification-engine.md
```diff
@@ -0,0 +1,238 @@
...
+- **Contribution + review pipeline** — every submitted decoder ships with positive and negative fixtures and an expected classification; CI schema-validates it, then runs it against the frozen evaluation corpus **across the supported SDK matrix**; lightweight acceptance via the Daml tooling SIG keeps the canonical set coherent.
+- **Public decoder corpus + evaluation harness** — the labeled ground-truth set the classifier is measured against. Labels come from **fault injection** (deterministic true labels — a deliberately missing controller has a known class) and **blind adjudication of real production rejections** (labeled without sight of the engine's output), split into a training set and a **frozen held-out set** contributors do not tune against. The harness computes the precision/coverage (risk-coverage) curve, the per-family confusion matrix, and confidence calibration; it is parameterised by SDK version so the same corpus replays against every supported generation.
```
daravep: Can you elaborate what a per-family confusion matrix is supposed to be? And what does the sentence "so the same corpus replays against every supported generation"?

proposals/canton-failure-classification-engine.md
```diff
@@ -0,0 +1,238 @@
...
+- **Integration contract** — a stable library interface plus a versioned JSON diagnosis schema, so DPM Trace, DPM Debug, the Transaction Profiler, CI runners, and IDE or ticketing integrations can call a measured classifier instead of each reimplementing failure heuristics.
```
daravep: Can you elaborate what this should mean?

daravep (commented):
The proposal appears to be rather convoluted. It does seem to miss that the error code inventory already exists but then also proposes a series of external tools that don't seem to make structurally sense.

proposals/canton-failure-classification-engine.md
infrasingularity: Fair question, and the file should have addressed it head-on — fixing that now. The error-code inventory and this engine sit at different layers, and the engine depends on the inventory rather than duplicating it.

The inventory defines what each code means and gives a generic resolution — one entry per code, the same text for every occurrence. But one code maps to many concrete root causes. A single authorization or contract-visibility rejection can be, in practice: a missing controller in the template design, an act-as/read-as claim missing from the token, a party hosted on a different participant than assumed, a divulgence gap, or package-vetting drift on the counterparty's node. The inventory entry can't distinguish these because it has no access to the specific rejection payload, the submitted command, the party set, or the package state. The engine takes exactly those inputs and returns which of the known causes produced this failure, with the specific next checks — measured for accuracy against a labeled test set.

Structurally, duplication is impossible: decoder matchers key on the official error codes and structured payload fields (2.1), so the inventory is an input dependency. We'll make that explicit — taxonomy families will cross-reference official error-code IDs, and every diagnosis will link back to the official entry for that code. Rough analogy: the inventory is the HTTP status registry; this is the tool that tells you why your request returned 403.

proposals/canton-failure-classification-engine.md
infrasingularity: Both are plain ideas dressed in bad language — rewording the file. Per-family confusion matrix: a table showing, for each failure family, where the engine's wrong answers land — e.g., how often a visibility failure gets mislabeled as an authorization failure. It's published so the committee can verify the headline precision number isn't hiding one weak family. "Replays against every supported generation": rejection payloads change shape across SDK releases, so every test case is tagged with the SDK version it came from and CI re-runs the full test set against each supported version. A new release that changes an error surface then shows up as a CI failure before it ever shows up as a wrong diagnosis for a user.

proposals/canton-failure-classification-engine.md
infrasingularity: These four bullets are one thing, not four tools — a single public repository and the machinery every serious open-source rules project has around it. In plain terms:

Knowledge base — the git repo holding the failure taxonomy and the decoder rules. Releases are checksummed and immutable, so every diagnosis can state exactly which rule-set version produced it. That's an audit requirement, not a product.

Contribution pipeline — the repo's CI and review process. A new rule arrives as a PR with examples it must classify correctly and examples it must not match; CI validates the format, runs it against the shared test set on each supported SDK version, and rejects it if it makes any existing category less accurate. The Daml tooling SIG signs off on inclusion.

Corpus + harness — the repo's test suite. Test cases come from two sources: deliberately induced failures where the true cause is known by construction (remove a controller from a template — the class is known), and real production rejections labeled by someone who hasn't seen the engine's output, so we're not grading our own homework. Part of the set is frozen and withheld so contributors can't tune rules to pass it. The "per-family confusion matrix" is just the table showing where wrong answers land — e.g. how often a visibility failure gets mislabeled as an authorization failure — so a headline accuracy number can't hide one weak category. "Replays against every supported generation" means each test case is tagged with the SDK version it came from and CI re-runs the whole set on every supported version, so an SDK release that changes error formats breaks the build instead of producing wrong diagnoses in the field.

Integration contract — the output format: a versioned JSON schema for the diagnosis plus a small library API, so DPM Trace, DPM Debug, the Profiler, or a CI runner can call the classifier and consume the result programmatically instead of re-implementing the rules.

proposals/canton-failure-classification-engine.md
daravep: Before I read into the details of your response, can you confirm that you are not in violation of https://github.com/canton-foundation/canton-dev-fund/blob/main/AI_POLICY.md?

Amit-fs: Yes, I have gone through it. If you have any questions or concerns, please let me know which part, and I can explain it in more detail.

hythloda:
Thank you for the proposal. We are declining this request at this time.

The team already has an active Development Fund grant, and we would first like to see the deliverables and outcomes from that work before considering an additional grant. Once the existing grant is completed and evaluated, a future proposal may be considered based on the quality and impact of that work.

This decision does not preclude a future submission.

PR #300, opened 2026-05-05
Proposal: Canton Event Stream — real-time ledger events via WebSocket…

anand-qasara:
**Proposal file:** /proposals/canton-event-stream.md

**Summary**
Canton Event Stream is a self-hostable, MIT-licensed Node.js service that connects to any Canton participant node and delivers classified, party-routed ledger events to application clients over WebSocket and SSE. It solves the missing push layer in the Canton ecosystem — the infrastructure between the raw JSON Ledger API stream and live UIs, deposit detectors, and reactive backends. The core implementation exists today inside Qasara's commercial Canton Gateway; this proposal extracts and open-sources it as a standalone package in 3 milestones over 6 weeks, funded at 300,000 CC.

**Checklist**

**Notes for Reviewers**
- The core ingester (Splice interface subscription, Redis pub/sub fanout, WebSocket + SSE endpoints) is already running in Qasara's DevNet environment — this is an extraction + open-sourcing grant, not a greenfield build.
- This proposal complements PQS (CIP-0100, approved 2026-04-23) rather than competing with it. PQS is the pull layer (SQL ODS); Canton Event Stream is the push layer (WebSocket/SSE). The Rationale section explains the coexistence design.
- Champion: Parth (parth@canton.foundation), Canton Foundation.
- Funding request: 300,000 CC across 3 milestones (120k / 100k / 80k).

### DIFF
--- a/proposals/canton-event-stream.md
+++ b/proposals/canton-event-stream.md
@@ -2,7 +2,7 @@
 **Status:** Submitted
 **Created:** 2026-05-05
 **Label:** canton-apis
-**Champion:** Parth <parth@canton.foundation>
+**Champion:** Jatin <jatin@canton.foundation>
 
 ---
 
--- a/proposals/canton-event-stream.md
+++ b/proposals/canton-event-stream.md
@@ -1,0 +1,326 @@
+**Author:** Anand <anand@qasara.ai>, Qasara
+**Status:** Submitted
+**Created:** 2026-05-05
+**Label:** canton-apis
+**Champion:** Parth <parth@canton.foundation>
+
+---
+
+# Canton Event Stream — Real-Time Ledger Events via WebSocket & SSE
+
+> Canton Development Fund Proposal
+> Funding request: 300,000 CC across 3 milestones (6 weeks)
+
+---
+
+## Abstract
+
+Canton Event Stream is an open-source, MIT-licensed, self-hostable Node.js service that connects to any Canton participant node and delivers classified, party-routed ledger events to application clients over WebSocket and Server-Sent Events (SSE). It fills the missing push layer in the Canton ecosystem — the infrastructure between the raw JSON Ledger API stream and the live UIs, deposit detectors, and reactive backends that real applications need. The core implementation already exists and runs against Canton DevNet inside Qasara's commercial Canton Gateway; this proposal extracts and open-sources it as a standalone package, complementing PQS (CIP-0100, the pull/SQL layer) on the same validator without overlap.
+
+---
+
+## Specification
+
+### 1. Objective
+
+Provide every Canton application team with a maintained, open-source push layer for ledger events — classified into named application events (`transfer.executed`, `transfer.rejected`, etc.), routed by party, and exposed over WebSocket and SSE so that browsers, mobile apps, and reactive backends can subscribe directly without each team rebuilding the same classification, fanout, and reconnection plumbing from scratch.
+
+The intended outcome: within 12 months of M1 release, building a real-time Canton dApp UI requires running a single Docker container alongside a validator — not designing and operating bespoke event-routing infrastructure.
+
+### 2. Implementation Mechanics
+
+**Architecture:**
+
+```
+Canton Participant Node
+     │
+     │  JSON Ledger API (WebSocket)
+     │  subscribeToUpdates({ interfaceIds: [...Splice CIPs...] })
+     ▼
+┌─────────────────────────────────┐
+│       Ledger Ingester           │
+│  - Splice interface subscription│
+│  - Event classification         │
+│  - Batch pub to Redis pipeline  │
+│  - Exponential backoff reconnect│
+└──────────────┬──────────────────┘
+               │  Redis Pub/Sub
+               │  channel: party:<partyId>
+               │  buffer:  eventbuf:<eventId> (5-min TTL)
+               ▼
+┌─────────────────────────────────┐
+│      Event Stream Server        │
+│  - WebSocket endpoint           │
+│  - SSE endpoint                 │
+│  - Multi-party subscription     │
+│  - Reconnect replay from buffer │
+│  - Token-based auth             │
+└──────────────┬──────────────────┘
+               │
+     ┌─────────┴──────────┐
+     ▼                    ▼
+WebSocket clients     SSE clients
+(browser dApps,    (server backends,
+ mobile apps)       monitoring tools)
+```
+
+**Workflow:**
+
+1. The Ledger Ingester connects to a Canton participant's JSON Ledger API and subscribes to ledger updates using Splice interface IDs (CIP-0056 token holdings + transfer instructions), with fallback to template-based subscription on nodes that do not yet support interface filtering.
+2. Raw DAML contract events are classified by parsing `templateId` and exercised choice names into human-readable typed events (`transfer.pending`, `transfer.executed`, `transfer.accepted`, `transfer.rejected`, `transfer.withdrawn`, `contract.created`, `contract.archived`).
+3. Events are routed by extracting `signatories`, `observers`, `witnessParties`, and `actingParties` across event types, then fanned out to subscribers via per-party Redis Pub/Sub channels.
+4. Events are buffered for 5 minutes to support reconnection replay via `lastEventId` — clients reconnect and request all buffered events since a given event ID without losing data.
+5. Application clients (browsers, mobile, server backends) connect to the Event Stream Server's WebSocket or SSE endpoint with a party-scoped token; they never connect to the participant node directly.
+6. The whole service ships as a single Docker image deployable alongside any Canton node, with Redis as its only persistent dependency.
+
+**Event format** — all events follow a consistent schema regardless of source:
+
+```json
+{
+  "eventId": "1220abc...def-uuid",
+  "partyId": "alice::1220...",
+  "type": "transfer.executed",
+  "timestamp": "2026-04-07T12:00:00.000Z",
+  "data": {
+    "contractId": "00abc...",
+    "templateId": "Splice.Amulet:Amulet",
+    "choice": "Transfer",
+    "consuming": true,
+    "updateId": "1220...",
+    "effectiveAt": "2026-04-07T12:00:00.000Z"
+  }
+}
+```
+
+**Event types:**
+
+| Type | Trigger |
+|---|---|
+| `transfer.pending` | TransferInstruction contract created |
+| `transfer.executed` | Transfer/Execute choice exercised |
+| `transfer.accepted` | Accept choice exercised |
+| `transfer.rejected` | Reject choice exercised |
+| `transfer.withdrawn` | Withdraw choice exercised |
+| `contract.created` | Any non-transfer contract created |
+| `contract.archived` | Any contract archived |
+
+**Splice interface support** — interface-based subscription rather than template-based, making the ingester forward-compatible with new CIP-0056 token implementations without code changes:
+
+- `#splice-api-token-holding-v1:Splice.Api.Token.HoldingV1:Holding`
+- `#splice-api-token-transfer-instruction-v1:Splice.Api.Token.TransferInstructionV1:TransferInstruction`
+
+**Stack:** TypeScript / Node.js, `@canton-network/wallet-sdk` for ledger access, Redis for fanout and replay buffer, single Docker image (~150 MB), Helm chart for Kubernetes deployment.
+
+### 3. Architectural Alignment
+
+**Ecosystem Infrastructure**: Real-time event delivery is foundational. Wallet UIs, exchange deposit detection, DeFi liquidation triggers, RFQ negotiation, and compliance monitoring all depend on reliable push notifications from the ledger.
+
+**Developer Experience**: Today every team solves this independently. A single, well-maintained open-source solution eliminates duplicate engineering and reduces the surface area for subtle bugs in each custom implementation.
+
+**Complementary to PQS (CIP-0100)**: PQS is the pull layer (offset-paged SQL queries against a materialized Postgres ODS). Canton Event Stream is the push layer. Production applications need both. PQS exposes no LISTEN/NOTIFY, no change feed, no subscription primitive — it is exclusively offset-paged query functions. (Detailed comparison in Rationale below.)
+
+**CIP-0056 alignment**: Interface-based subscription means the ingester works with any CIP-0056 compliant token without modification, supporting the Canton multi-asset future.
+
+**Canton 3.5 alignment**: The compatibility matrix (M3) targets Canton 3.5+ explicitly, in line with the mid-June 2026 protocol upgrade.
+
+### 4. Backward Compatibility
+
+No backward compatibility impact. Canton Event Stream is additive infrastructure that subscribes to the existing JSON Ledger API; it does not modify Canton, Splice, or any existing client interface. Validators electing to deploy it run a new sidecar Docker container — no changes to their participant node, no schema migrations, no client breakage.
+
+---
+
+## Milestones and Deliverables
+
+### Milestone 1: Standalone Package + Replay
+- **Estimated Delivery:** Week 2
+- **Focus:** Extract core ingester from Qasara's commercial gateway into a self-contained MIT-licensed package; add reconnection replay.
+- **Deliverables / Value Metrics:**
+  - npm package decoupled from Qasara's auth, API key, and database systems
+  - Standalone configuration via environment variables only (no Prisma, no external DB dependency)
+  - Multi-auth support: sandbox HMAC, DevNet OAuth2 ClientCredentials, configurable per deployment
+  - `lastEventId` replay on reconnect (5-minute buffer window)
+  - Docker image published to GitHub Container Registry
+  - Unit tests (≥80% coverage on ingester and event classification logic)
+  - GitHub Actions CI: lint, test, Docker build on every PR
+
+### Milestone 2: Integration Tests + AsyncAPI Spec
+- **Estimated Delivery:** Week 4
+- **Focus:** Production-grade test coverage and developer-facing API documentation.
+- **Deliverables / Value Metrics:**
+  - End-to-end integration test suite against Canton Quickstart in CI:
+    - WebSocket connect, subscribe, receive events, disconnect
+    - SSE connect, receive events, disconnect
+    - Stream token issue/use-once/reject-second-use
+    - `lastEventId` replay after disconnect
+    - Invalid token rejection
+    - Multi-party subscription routing
+  - AsyncAPI 3.0 specification covering both WebSocket and SSE endpoints, published and rendered as documentation
+  - `docker-compose.yml` for local development (redis + canton-event-stream + Canton Quickstart)
+  - Published developer quickstart guide: zero-to-events in under 10 minutes
+
+### Milestone 3: Production Hardening + Community Onboarding
+- **Estimated Delivery:** Week 6
+- **Focus:** Make the service safely runnable on production validators and ready for community adoption.
+- **Deliverables / Value Metrics:**
+  - Helm chart for Kubernetes deployment with configurable replicas, resource limits, and Redis connection settings
+  - Prometheus metrics endpoint (`/metrics`): active connections, events/sec, Redis publish latency, reconnect count
+  - Grafana dashboard JSON
+  - Structured JSON logging with configurable log level
+  - Compatibility matrix: Canton Quickstart, Canton 3.5+ DevNet, MainNet validator (Canton 3.5+), Splice 0.5.x and 0.6.x
+  - Security checklist: stream token entropy, WebSocket connection limits, Redis ACL configuration guide
+  - Contributor guide and governance documentation
+  - 12-month maintenance commitment: security patches, Splice/SDK version compatibility updates, issue triage
+
+---
+
+## Acceptance Criteria
+
+The Tech & Ops Committee will evaluate completion against ecosystem-value metrics for each milestone, supported by underlying artifact evidence.
+
+### Milestone 1 — Adoption-readiness
+- ≥1 external developer (non-Qasara) successfully connects to Canton Quickstart and receives classified events using only the published README — no Qasara-specific guidance required
+- Reconnection replay verified end-to-end: client disconnects mid-stream and recovers all events within the 5-minute buffer using `lastEventId`
+- Package and Docker image publicly available on npm and GHCR; install from a fresh machine reaches a working event stream in a single `docker compose up`
+- Unit test suite passes in CI
+
+### Milestone 2 — Developer onboarding velocity
+- ≥3 external teams (validators or app builders) confirm successful local setup using the quickstart guide and provide written feedback
+- AsyncAPI spec published and renderable in standard tooling (e.g., AsyncAPI Studio)
+- An external tester with no prior Canton experience reaches a working event stream within 10 minutes of opening the guide
+- Integration test suite passes in CI against Canton Quickstart
+
+### Milestone 3 — Production deployment readiness
+- ≥2 external production-track deployments running canton-event-stream alongside their validator (DevNet or MainNet), OR ≥1 Canton Foundation-recognized reference deployment
+- Helm chart deployed and verified on a real Kubernetes cluster (not a single-node test rig)
+- Compatibility matrix published with passing test results across all listed environments
+- Security checklist reviewed and signed off by a Canton Foundation Tech & Ops Committee member
+- ≥1 external contributor PR successfully merged following the contributor guide
+
+---
+
+## Funding
+
+**Total Funding Request:** 300,000 CC
+
+### Payment Breakdown by Milestone
+- Milestone 1 (Standalone Package + Replay): **120,000 CC** upon committee acceptance
+- Milestone 2 (Integration Tests + AsyncAPI Spec): **100,000 CC** upon committee acceptance
+- Milestone 3 (Production Hardening + Community Onboarding): **80,000 CC** upon final release and acceptance
+
+### Volatility Stipulation
+
+Project duration is **under 6 months** (6 weeks planned). Should the project timeline extend beyond 6 months due to Committee-requested scope changes, any remaining milestones must be renegotiated to account for significant USD/CC price volatility.
+
+---
+
+## Co-Marketing
+
+Upon release, Qasara will collaborate with the Canton Foundation on:
+
+- **Announcement coordination** — joint blog post and Canton Foundation channel announcements at M1 (npm release) and M3 (production-ready release)
+- **Technical case study** — a published deep-dive on the architecture, the PQS-vs-push-layer design rationale, and lessons from running this in production at Qasara
+- **Reference integration** — make Qasara's Canton Gateway integration with canton-event-stream a public reference example for other dApp builders
+
+---
+
+## Motivation
+
+Real-time event delivery is foundational infrastructure for the Canton dApp ecosystem. Wallet UIs, exchange deposit detection, DeFi liquidation triggers, RFQ negotiation, and compliance monitoring all depend on reliable push notifications from the ledger. Today every team solves this independently — duplicating engineering effort, multiplying the surface area for subtle bugs, and slowing dApp time-to-market.
+
+**The raw JSON Ledger API is not the answer for application clients.** It exposes a WebSocket stream — but what that stream delivers is raw DAML contract events:
+
+```json
+{
+  "update": {
+    "value": {
+      "updateId": "1220abc...",
+      "effectiveAt": "2026-04-08T...",
+      "events": [
+        {
+          "contractId": "00abc...",
+          "templateId": "#splice-amulet-0.1.9:Splice.Amulet:Amulet",
+          "signatories": ["alice::1220..."],
+          "observers": [],
+          "createdAt": "2026-04-08T...",
+          "createdEventBlob": "CgYI..."
+        }
+      ]
+    }
+  }
+}
+```
+
+To turn this into something an application can act on, every team must build event classification, party routing, multi-subscriber fanout, reconnection-with-replay, and a client-facing WebSocket/SSE server (so browsers and mobile apps don't connect to the participant node directly). This abstraction layer does not exist as a shared, maintained open-source component anywhere in the Canton ecosystem.
+
+**Ecosystem portion benefited:** We estimate **~70% of new Canton dApp development happens on Node.js / TypeScript stacks** — the Canton Wallet SDK is TypeScript-native, and the dApp/fintech developer mainstream is Node.js. Of those teams, **all that need real-time event delivery are addressable** — every wallet UI, every exchange integration, every reactive backend. We conservatively expect **≥10 production integrations within 12 months** of M3 release, and a measurable reduction in time-to-first-event for new Canton applications. The remaining ~30% (JVM-stack teams) can still consume the WebSocket/SSE endpoints from any language; they simply gain less from the TypeScript-native implementation.
+
+**Strategic importance:** Approving PQS (CIP-0100) without a complementary push layer leaves every Canton application team to rebuild that push layer independently — exactly the duplication a foundational push component eliminates. PQS makes ledger state queryable and historically auditable; Canton Event Stream makes ledger events deliverable to live UIs and reactive backends. Production applications need both layers.
+
+---
+
+## Rationale
+
+### Why a new component, not an extension of PQS
+
+CIP-0100 (PR #67, merged 2026-04-23) approved Digital Asset's grant to open-source PQS — the Participant Query Store — under Apache 2.0. PQS and Canton Event Stream serve different layers of the same stack and cannot be folded into one another:
+
+| Dimension | PQS (offset-paged SQL ODS) | Canton Event Stream (push) |
+|---|---|---|
+| **Consumer-facing interface** | SQL API: `active()`, `creates()`, `archives()`, `exercises()` | WebSocket and Server-Sent Events |
+| **Wire protocol** | PostgreSQL JDBC | WebSocket / HTTP SSE |
+| **Client topology** | Server-side only (browsers cannot speak Postgres) | Browser, mobile, and server clients |
+| **Authentication** | PostgreSQL roles | App-layer party-scoped tokens |
+| **Event shape** | Decoded contracts in JSONB; clients implement classification | Pre-classified app events (`transfer.executed`, `transfer.rejected`, etc.) |
+| **Hot-path latency** | Ledger API → PQS write → Postgres → poll | Ledger API → Redis pub/sub → WebSocket push (no DB write in the hot path) |
+| **Reconnection model** | Client tracks Postgres offset; bespoke replay logic | `lastEventId` against a 5-minute event buffer, designed for browser reconnect |
+| **Footprint** | JVM (4 GB / 4 cores min.) plus Postgres (8 GB / 8 cores min.) | Node.js (~150 MB image), Redis only |
+
+The [PQS SQL API documentation](https://docs.digitalasset.com/build/3.5/component-howtos/pqs/references/sql-api.html) confirms the consumer-facing interface is exclusively offset-paged query functions — no LISTEN/NOTIFY, no change feed, no subscription primitive. Adding push to PQS would require building exactly this proposal inside a JVM project, which forces a JVM dependency on every dApp consumer (the very dependency many builders are trying to avoid).
+
+### Why TypeScript / Node.js
+
+Existing Canton event indexing work in the ecosystem is JVM-based (Scala). The Canton SDK itself is TypeScript-native (`@canton-network/wallet-sdk`). For teams already running a Node.js application stack — the majority of dApp and fintech developers — a TypeScript implementation:
+
+- Eliminates a JVM dependency from their infrastructure
+- Uses the same SDK they are already familiar with
+- Produces a smaller Docker image (~150 MB vs ~400 MB+ JVM)
+- Shares package ecosystem with the Canton Wallet SDK
+
+JVM-stack consumers are not excluded — the WebSocket and SSE endpoints are language-neutral wire protocols, so any client (Java, Python, Rust, Go) connects without a Node.js runtime on its side.
+
+### Why interface-based subscription
+
+Building this on top of template-based subscription would lock the ecosystem into per-token rewrites every time a new asset launches; building on Splice interfaces (CIP-0056) means the ingester works with any compliant token without modification, supporting the Canton multi-asset future. The fallback to empty-template subscription preserves compatibility with nodes that do not yet support interface filtering.
+
+### Coexistence on the same validator
+
+PQS and Canton Event Stream are independent of one another and can run on the same validator without interaction. Canton Event Stream subscribes to the Canton Ledger API directly and uses Redis as its only persistent state — it does not depend on Postgres or share any infrastructure with PQS. Validators electing to run both gain queryable historical state from PQS and live event delivery from Canton Event Stream without integration work between the two.
+
+### Alternatives considered
+
+- **Recommend each team builds their own push layer over the JSON Ledger API.** Status quo; results in fragmented, inconsistent event semantics, duplicated bugs in classification and reconnection logic, and slow dApp time-to-market. Rejected as the explicit problem this proposal solves.
+- **Add push to PQS.** Forces a JVM dependency on every consumer and changes PQS's architectural identity from a query store into a hybrid push+query system. Better served by an independent, lightweight push component.
+- **Standardize on the gRPC streaming Ledger API.** gRPC streaming is excellent for server-to-server but not consumable by browsers or mobile clients without a translation proxy — exactly the proxy this proposal builds, only the WebSocket/SSE-facing version of it is broadly consumable.
+
+---
+
+## Maintenance Plan
+
+- MIT license, hosted at `github.com/qasara/canton-event-stream`
+- 12-month post-launch maintenance: security patches, Splice/SDK version compatibility updates
+- Semantic versioning with a documented breaking-change policy
+- Issues triaged weekly; critical fixes within 48 hours
+- Compatibility tested against each major Splice and Canton release
+- After 12 months: proposal to transfer stewardship to Canton Foundation or a community SIG if formation is possible
+
+---
+
+## Notes for Reviewers
+
+1. **Core implementation exists today.** This is not a design proposal. The ledger ingester (Splice interface subscription, exponential backoff reconnect), Redis Pub/Sub fanout with batch pipeline, WebSocket endpoint, and SSE endpoint are all implemented and running in our DevNet environment as part of the Qasara Canton Gateway. The milestones deliver: extracting this into a standalone open-source package, decoupling it from Qasara's auth and API key system, adding the `lastEventId` replay query on reconnect, and adding production hardening (Helm chart, metrics, standalone Docker image). The hard engineering is done — what remains is packaging, decoupling, and documentation.
+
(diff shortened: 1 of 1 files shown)


Denend:
Quick question - how does canton event stream position itself relative to Cantons native push layer (json ledger api v2 ws channels via AsyncAPI, plus grpc UpdateService.GetUpdates server-streaming)? Would help to see the delta laid out

  Refs:
  - https://github.com/digital-asset/canton/blob/main/community/ledger/ledger-json-api/src/test/resources/json-api-docs/asyncapi.yaml
  - https://github.com/digital-asset/canton/blob/main/docs-open/src/sphinx/sdk/tutorials/json-api/canton_and_the_json_ledger_api_ts_websocket.rst

Denend:
@anand-qasara

anand-qasara:
> Quick question - how does canton event stream position itself relative to Cantons native push layer (json ledger api v2 ws channels via AsyncAPI, plus grpc UpdateService.GetUpdates server-streaming)? Would help to see the delta laid out
> 
> Refs:
> 
> * https://github.com/digital-asset/canton/blob/main/community/ledger/ledger-json-api/src/test/resources/json-api-docs/asyncapi.yaml
> * https://github.com/digital-asset/canton/blob/main/docs-open/src/sphinx/sdk/tutorials/json-api/canton_and_the_json_ledger_api_ts_websocket.rst



> @anand-qasara

Good question — and an important distinction. Let me lay out the delta precisely.

Canton's native push layer — gRPC UpdateService.GetUpdates and the AsyncAPI-documented JSON Ledger API WebSocket channels (/v2/updates et al.) — is the transport, and it's complete: multi-party filtering (filtersByParty), offset/checkpoint resume, and LEDGER_EFFECTS choice data are all first-class. Canton Event Stream is not an alternative transport. It consumes the native JSON WS upstream and sits above it as the application-edge layer.

The delta is the part the transport intentionally doesn't provide, and the core of it is privacy:

When a client talks to the participant directly, the node enforces per-party authorization — the token's read-rights bound which parties it can put in filtersByParty, so a consumer physically cannot subscribe to a party it isn't entitled to. That's exactly right, but it's a per-client model: every browser/microservice opens its own node subscription, holds a Ledger API token, and receives raw Created/Archived/Exercised primitives it has to classify itself.

The moment you add a fan-out tier to serve many clients from one upstream subscription — which you must, to avoid exposing the participant to every client and handing every client a ledger token — the node's per-party authorization boundary is removed from the downstream path. Something has to re-establish it, or one party's events leak to another. A generic WebSocket fan-out (Centrifugo, socket.io + Redis) can't: it has no concept of a Canton party or of who is entitled to see whose events. Re-enforcing per-party authorization at the streaming edge is the Canton-specific, security-critical work that makes a purpose-built event gateway necessary rather than off-the-shelf pub/sub — and it's the central deliverable of this proposal (M1), not an afterthought.

So the three deltas over the native transport are:

Per-party authorization at the edge — preserving Canton's privacy guarantee across the client-decoupling boundary (the hard part; generic infra gets this wrong).
Fan-in/fan-out — one upstream node subscription serving many clients, so the participant isn't exposed to or loaded by every client, and clients never hold a ledger token.
CIP-0056 event classification — mapping raw Created/Exercised(choice, consuming) into the stable token-standard vocabulary (transfer.executed, …) so every team doesn't re-implement and re-maintain the same Splice-interface→event mapping.
It composes with the native layer rather than replacing it — the same relationship the proposal draws with PQS on the pull side (native primitive below, shared application-layer infrastructure above). It also only ever delivers events for parties an operator is authorized to read; by Canton's authorization model it can't expose third-party private events.

I've updated the proposal to make this explicit: see the new "Relationship to Canton's native push layer (JSON WS and gRPC streaming)" section under Rationale, the per-party authorization deliverable in Milestone 1, and the reframed Abstract/Objective. Thanks for pushing on this — it sharpened the positioning.

proposals/canton-event-stream.md
```diff
@@ -0,0 +1,342 @@
...
+**Status:** Submitted
+**Created:** 2026-05-05
+**Label:** canton-apis
+**Champion:** 
+
+---
+
+# Canton Event Stream — Real-Time Ledger Events via WebSocket & SSE
+
+> Canton Development Fund Proposal
```
hrischuk-da, line 11: I have reviewed the proposal.  I think the approach has merit but I find this proposal is lacking and needs more updates.  I could see this as being useful for scaling up the LAPI event stream to more clients but there are remaining questions.

The larger gaps to close are:
- It does not support multi-hosted parties and party based HA (future) whereby it can switch to a different validator LAPI if the one it is connected to fails.  Multi-hosted support should be the default.
- The ordering mechanism is not sufficient.  It would need to use RecordTime as its ordering metric.  This would also allow a subscriber to say from when it wants to receive events.
- It is not clear what the fan-in-fan-out scaling benefit is because each user/part would make a subscription for itself using its own JWT authorized to its own party.  Naively, this looks to me like a 1-1 passthrough authorization which will not benefit scalability but I may be missing something.
- There is a very narrow focus on what it subscribes to and there is no explanation for this.  The list is too narrow for a general purpose solution and it should be able to be added to which brings me to ...
- Creating a CIP to standardize the ledger event to subscribe event would be appropriate.
- Needs to support 'unassign' and 'reassign' events for multi-synchronizer apps.

Happy to discuss more.

anand-qasara: Thanks — sharp review, and it genuinely improves the design. We agree with most of it, want to restructure into two phases, and would value a quick call to align before we revise.

- RecordTime ordering + "from when" resume (2): agreed — we'll make RecordTime the ordering/resume     key, with an offset-checkpoint index to translate a requested time into a resume point.

- Configurable subscriptions (4): agreed — the token-interface default reflects the code's origin, not a design limit; we'll make the subscribed interfaces/templates configurable with pluggable classification.

- Unassign/assign events (6): agreed — we'll classify reassignment so multi-synchronizer apps don't misread a contract moving synchronizers as archive+create.

- CIP (5): strong yes — we'll author a CIP for the event-selection and classified-event model, with this service as the reference implementation; we'd welcome your steer on the right SIG and path.

- Multi-hosted + party HA (1): agreed on posture — RecordTime is the enabler (participant-independent key, failover, dedup). It can only be exercised on a multi-participant network, so we'd scope it (with reassignment) as a second phase on a multi-synchronizer/multi-participant testbed.

- On fan-in/fan-out (3) — fair challenge. You're right that a per-user party self-subscribing with its own JWT is 1-1, with no subscription-dedup benefit. The benefit is in the operator-hosted model: one credential authorized for many parties opens one (or a few) multi-party subscriptions, fanned out to that operator's many downstream consumers — so downstream client connections decouple from upstream node subscriptions, and the participant is shielded from per-client connections each holding a ledger token. In the per-user-owned-party model the value is connection multiplexing, classification, and resumable ordering — not subscription dedup. We'll state that deployment model explicitly.

Net: Phase 1 (demonstrable now) — RecordTime resume, configurable subscriptions, classification, per-party authorization, CIP draft. Phase 2 (needs a multi-participant/multi-synchronizer testbed) — multi-hosted failover and reassignment. Could we grab 30 minutes to align on the CIP path and the phase boundary before we push the revision? 

Thanks again.

Denend:
The M1 acceptance criterion requires demonstrating that a client requesting an unauthorized party is rejected at subscribe time. How does the Event Stream Server validate downstream client credentials - who issues stream tokens, what party claims do they encode, and how does the server verify a client's party entitlement at subscribe time?

anand-qasara:
> The M1 acceptance criterion requires demonstrating that a client requesting an unauthorized party is rejected at subscribe time. How does the Event Stream Server validate downstream client credentials - who issues stream tokens, what party claims do they encode, and how does the server verify a client's party entitlement at subscribe time?

@Denend 

Good question — this is the exact mechanism Milestone 1 builds. Here's how it works.

**Who issues the stream tokens.** Whoever already knows which user maps to which Canton party — the operator running the service (e.g. an exchange that knows "this logged-in user = party alice"). The Event Stream Server doesn't decide who owns which party; it only _trusts_ tokens signed by an issuer it's configured to accept (via a shared key or a JWKS URL). This is the same trust model the participant already uses — it trusts the `actAs` claims in an IdP-signed JWT — just applied one hop further out, at the streaming edge.

Two deployment shapes: (a) the service issues the token itself after authenticating the client and looking up its parties from an operator-supplied mapping, or (b) the operator's existing identity system issues the token and the service just verifies it. Making that verify step pluggable — so any operator can use their own identity system — is the net-new open-source work. (In our commercial gateway today it's shape (a), tied to our API-key system.)

**What's in the token.** An explicit list of the party IDs the holder is allowed to stream, plus standard `iss`/`aud`/`exp` and a single-use id. The allowed parties are written into the signed token by the issuer — the server never guesses them. Tokens are short-lived (~60s) and single-use.

**How the server checks, at subscribe time.** When a client connects and asks to stream a set of parties, the server before opening any channel:

1. Verifies the token (signature, audience, not expired, not already used).
2. Reads the allowed-party list from the token.
3. Checks every requested party is on that list. If any isn't, the subscribe is rejected and no channel opens.

So the per-party boundary the node enforces on a direct subscription is re-created at the edge: the token's party list bounds what a client can stream. That's exactly what the M1 acceptance test shows — ask for a party your token doesn't include, and you're rejected.

Optionally (configurable), the service can also double-check with the participant that the token's issuer really holds read rights for that party — so even a mis-issued token can't grant more than the node itself would allow.
I can add this as an "Authorization model" subsection in the proposal so it's written down explicitly.

Denend:
@anand-qasara thanks, that covers it

Erlannn7:
This is a practical addition to the Canton ecosystem. Having an open-source event streaming layer for WebSocket and SSE could make it much easier to build responsive dApps and real-time user experiences. I also appreciate the clear positioning as a complement to PQS rather than a replacement.

Noosphere-314:
We're Woof ([woof-software](https://github.com/woof-software)) - currently prepping two Dev Fund proposals ([write-up](https://forum.canton.network/t/woof-two-dev-fund-proposals-evm-side-compliance-risk-infrastructure-for-canton-seeking-feedback-champion/8796)): an EVM-side compliance SDK for Zenith and open-source risk primitives. Flagging interest because both would otherwise have to build this exact layer ourselves - our TS middleware needs live per-party compliance and settlement events for frontends, and our risk dashboard rebuilds state from classified events. If this lands we'd look at consuming it instead of rolling our own.

And we're not the only ones rolling our own, which is really the argument for funding it once. Hacken (#302) promises alerts "within 60 seconds of a triggering event," Zpoken (#138) is bolting pg_notify + GraphQL subscriptions onto their indexer for "within 2s" push, and #51 (Participant Event Gateway) solves the same participant-local, party-scoped delivery problem over webhooks. The 2026 DevEx survey lists real-time streaming as a gap people patch individually. Worth linking #51 from here - same problem, different transport, so it reads as demand rather than overlap.

On the native-vs-this question from earlier: we went through the JSON Ledger API v2 AsyncAPI spec and gRPC UpdateService ourselves, and the delta holds. Native gives you per-party-filtered streams with offset resume and stakeholder auth, but every client connects with the participant's own JWT and minds its own offsets - and the docs themselves say to keep long-lived streams to a minimum (WS caps around 2h). Once you fan one upstream subscription out to many clients, so they aren't each holding a ledger token, the participant's per-party check leaves the path, and a plain WS broker has no notion of a Canton party to put it back. Re-doing party auth at the edge, classifying raw Created/Exercised into transfer.executed, and offering SSE at all (native is WS/gRPC only) is a genuine layer. Push-here vs pull-from-PQS is the right split.

Three things we'd want pinned down before building on it, coming at it as a compliance/settlement consumer:

1. Version the taxonomy against the token-standard interfaces. The default profile maps HoldingV1 / TransferInstructionV1, and the event JSON has no schema-version field - only the package is semver'd. splice is shipping v1 and v2 interfaces in parallel right now (there's a weekly-reset V2 DevNet), so anything wired to today's shapes breaks quietly when instruments move to v2. A schemaVersion on the event, plus a stated policy for mapping across interface versions, saves every consumer from patching that later.

2. Say what the delivery guarantee is, and what happens at the resume edge. You've got RecordTime ordering and resume-within-a-window, but not the guarantee class (at-least-once? exactly-once?), how a client dedupes across a reconnect, or what a resume request older than the window returns. For deposit detection and compliance state that's not academic - a silent gap is a missed deposit, a dup is a double count. Simplest fix: declare at-least-once, give consumers a stable idempotency key (eventId?), and make an out-of-window resume a hard error so they fall back to PQS instead of assuming continuity.

3. Spell out how the token rides on the wire, especially for SSE. The short-lived single-use party token is a good design, but browser EventSource can't set headers, which tends to push the SSE token onto the query string - where it lands in proxy and access logs. Worth documenting the transport, and for SSE using something that keeps it out of the URL (short-lived cookie, or a POST to establish the stream).

Overall - real gap, the positioning against native/PQS checks out, and the thread's already sharpened it a lot. The above is hardening, not objection. Glad to dig into the compliance-event side of the taxonomy later if it's useful.

anand-qasara:
@Noosphere-314 
Thank you for the great feedback. It's a real demand signal and a proper hardening pass in one, and all three points are going straight into the design
 
Versioning the taxonomy — agreed. There's no schema-version field today and there should be. I'll put a schemaVersion on the event plus a clear policy for mapping across interface versions. V2: CIP-0112 adds a first-class EventLog_HoldingsChange event, so the right move is to version the taxonomy against the interface — infer-from-choice on V1, read-the-EventLog on V2 — and let schemaVersion tell the consumer which they've got. Then an instrument moving to v2 flips a field instead of silently breaking you.

Delivery guarantee — agreed, and you're right this can't stay implicit; for deposit/compliance state a silent gap is a missed deposit and a dup is a double count. So: at-least-once, a stable idempotency key, and an out-of-window resume that fails hard so you fall back to PQS instead of assuming continuity. Today's eventId has a random UUID in it, so it isn't stable across a re-ingest; it needs to be deterministic (updateId + event position). That would've bitten everyone.

Token on the wire — yep, the SSE-token-in-the-query-string thing is a genuine footgun (EventSource can't set headers, so it ends up in proxy logs). We'll document transport properly: WS in the Sec-WebSocket-Protocol subprotocol like the native API, and SSE via a short-lived cookie or a POST to open the stream — out of the URL either way.

Thanks again

anand-qasara:
@Freakzyyy45 Thanks, really appreciate it.

Jatinp26:
@anand-qasara it states that Qasara operates a hosted version as part of their commercial Canton Gateway. The OSS version excludes API key management, rate limiting, and billing remain commercial. As dev fund is to fund public good opensource products, I am not sure how that fits in with Qasara and how much of this money is dedicated to your commercial offering and what would OSS would offer if it doesn't have API Key management, need more transparency in it.

anand-qasara:
@Jatinp26 

Thanks Jatin, this is the right thing to push on.

To be clear on the money: the whole grant goes into the public MIT repo. None of it funds the hosted service — we just run the same open-source package ourselves. There's no private fork of the actual streaming engine.

On what you get without our API-key system: it's a full self-hostable service, not a stripped one. The ingester, the event classification, the per-party privacy enforcement at the edge (which is the part that's easy to get wrong, and is squarely OSS), fanout, WebSocket/SSE, reconnect, Helm charts — all public. It authenticates connections too; it validates a client's JWT/token, and it's decoupled from our own auth and DB so you plug in your own IdP. What's not in it is the SaaS billing layer — issuing and metering keys for paying customers. That's the Grafana OSS vs Grafana Cloud line, and it's where our commercial side actually sits.

On rate limiting — you're right, that's a wording error. Basic connection/backpressure limits belong in the OSS and I'll move them there explicitly; only per-plan tiered limits are commercial.

We run this in production ourselves, which is what keeps it maintained past the grant. And the incentive runs the same way: the more teams building live Canton apps, the bigger the market for our hosted API — keeping the engine closed would work against us.

I'll push a corrected version with an explicit OSS-vs-hosted table so none of this is left to interpretation.

proposals/canton-event-stream.md
```diff
@@ -0,0 +1,387 @@
...
+
+---
+
+## Abstract
+
+Canton Event Stream is an open-source, MIT-licensed, self-hostable Node.js service that connects to any Canton participant node and delivers classified, party-authorized ledger events to application clients over WebSocket and Server-Sent Events (SSE). Canton already provides a native push *transport* (the JSON Ledger API WebSocket channels and gRPC `UpdateService.GetUpdates`); Canton Event Stream is the **application-edge layer above it** — it takes one upstream subscription, re-enforces Canton's per-party visibility at the streaming edge, classifies raw events into a stable application vocabulary, and fans them out to many clients without each client connecting to the participant or holding a ledger token. The transport and event-classification core already run against Canton DevNet inside Qasara's commercial Canton Gateway; this proposal open-sources that core as a standalone package and builds the net-new pieces the ecosystem needs — per-party authorization at the streaming edge, RecordTime ordering, and configurable subscriptions — complementing PQS (CIP-0100, the pull/SQL layer) on the same validator without overlap.
```
hrischuk-da: Can we change this to Apache 2?

hrischuk-da: Can you add that your commercial offering is an infrastructure offering that will run the same OSS code as is developed here?

proposals/canton-event-stream.md
```diff
@@ -0,0 +1,387 @@
...
+
+---
+
+## Specification
+
+### 1. Objective
+
+Provide every Canton application team with a maintained, open-source application-edge layer over Canton's native push transport — classifying ledger events into named application events (`transfer.executed`, `transfer.rejected`, etc.), re-enforcing per-party authorization at the streaming edge, ordering and resuming by RecordTime, and exposing the result over WebSocket and SSE so that browsers, mobile apps, and reactive backends can subscribe without each team rebuilding the same classification, authorization, ordering, fanout, and reconnection plumbing from scratch — and without each client connecting directly to the participant. The set of events subscribed to is operator-configurable, not fixed to one asset family.
+
+The intended outcome: within 12 months of M1 release, building a real-time Canton dApp UI requires running a single Docker container alongside a validator — not designing and operating bespoke event-routing infrastructure.
```
hrischuk-da: This is a pretty strong claim.  Are you willing to bet all of the funding on that claim?  If not, I suggest weakening it to what can be achieved.

proposals/canton-event-stream.md
```diff
@@ -0,0 +1,387 @@
...
+
+This proposal funds **Phase 1** — the application-edge layer that is fully demonstrable on infrastructure available today (a single participant). **Phase 2** — multi-hosted high availability (failover across participant LAPI endpoints) and multi-synchronizer reassignment events — is scoped as a follow-on grant because it can only be built and demonstrated against a multi-participant / multi-synchronizer testbed. The Phase-1 architecture is designed so Phase 2 is additive, not a rewrite: RecordTime ordering (Phase 1) is the participant-independent key that multi-hosted failover and dedup (Phase 2) depend on.
```
hrischuk-da, line 31: You will also need to deal with deduplicating the same event from multiple participants because of multi-hosting of parties.

proposals/canton-event-stream.md
```diff
@@ -0,0 +1,387 @@
...
+```
+Canton Participant Node(s)
+   Phase 1: one LAPI endpoint
+   Phase 2: multiple participants hosting the same party set,
+            RecordTime-based failover + dedup across endpoints
+     │
+     │  JSON Ledger API (WebSocket)
+     │  configurable interface / template subscription
+     ▼
+┌──────────────────────────────────────┐
```
hrischuk-da, line 46: doesn't htis figure needs to show one or two clients that adjust their subscriptions?

proposals/canton-event-stream.md
```diff
@@ -0,0 +1,387 @@
...
+     ┌─────────┴──────────┐
+     ▼                    ▼
+WebSocket clients     SSE clients
+(browser dApps,    (server backends,
+ mobile apps)       monitoring tools)
+```
+
+**Workflow:**
+
+1. The Ledger Ingester connects to a Canton participant's JSON Ledger API and subscribes to a **configurable set of interfaces/templates** — operators declare what to subscribe to, with a pluggable classification mapping. The CIP-0056 token interfaces (holdings + transfer instructions) ship as the **default profile**, with fallback to template-based subscription on nodes that do not yet support interface filtering; the mechanism is general-purpose, not fixed to token assets.
```
hrischuk-da: Probably need to include Token Standard v2 since it is running live in DevNet.

proposals/canton-event-stream.md
```diff
@@ -0,0 +1,387 @@
...
+4. Events are ordered by **RecordTime** — Canton's participant-independent record timestamp, present on every update. Subscribers resume from a requested RecordTime ("send me everything since T") within a **configurable recent retention window** — a live-reconnect aid on the order of minutes to hours, **not a historical store**; deep historical replay is PQS's role, not this service's. The service maintains a RecordTime→offset index over that window to translate a requested time into a participant offset to resume from. (Choosing RecordTime over a participant-local offset is also what makes Phase-2 multi-hosted failover and dedup correct — see Phasing.)
+5. Application clients (browsers, mobile, server backends) connect to the Event Stream Server's WebSocket or SSE endpoint with a party-scoped token; they never connect to the participant node directly.
+6. The whole service ships as a single Docker image deployable alongside any Canton node, with Redis as its only persistent dependency.
+
+**Event format** — all events follow a consistent schema regardless of source:
+
+```json
+{
+  "eventId": "1220abc...def-uuid",
```
hrischuk-da: Isn't this missing important information from the token standard v1 and v2?   Is that buried somewhere in these fields?

proposals/canton-event-stream.md
```diff
@@ -0,0 +1,387 @@
...
+| `contract.created` | Any non-transfer contract created |
+| `contract.archived` | Any contract archived |
+| `contract.unassigned` | Contract unassigned from a synchronizer (reassignment out) — *Phase 2* |
+| `contract.assigned` | Contract assigned to a synchronizer (reassignment in) — *Phase 2* |
+
+The event vocabulary above is the default profile; because the subscription set and classification mapping are configurable, operators can extend it to their own templates. The `unassigned`/`assigned` (reassignment) events matter for multi-synchronizer apps — without them, a contract moving between synchronizers is misread downstream as archive-then-create. They require ≥2 synchronizers to exercise and are delivered in **Phase 2** (see Phasing).
+
+**Subscription mechanism** — the ingester subscribes by **interface ID where available** (forward-compatible: a new template implementing an existing interface is picked up with no code change) and falls back to **template-based** subscription on nodes that don't yet support interface filtering. The subscription set is **operator-configurable**; the CIP-0056 token-standard interfaces ship as the default profile:
```
hrischuk-da: Please explain what a 'default profile' looks like and what is in it.

proposals/canton-event-stream.md
```diff
@@ -0,0 +1,387 @@
...
+
+- `#splice-api-token-holding-v1:Splice.Api.Token.HoldingV1:Holding`
+- `#splice-api-token-transfer-instruction-v1:Splice.Api.Token.TransferInstructionV1:TransferInstruction`
+
+Operators can add their own interfaces or templates to the set — the mechanism is general-purpose, not token-specific.
+
+**Stack:** TypeScript / Node.js, `@canton-network/wallet-sdk` for ledger access, Redis for fanout and replay buffer, single Docker image (~150 MB), Helm chart for Kubernetes deployment.
```
hrischuk-da, line 226: Can you provide a block diagram that explains what this solution will look like for an end-to-end solution?

proposals/canton-event-stream.md
```diff
@@ -0,0 +1,387 @@
...
+
+## Milestones and Deliverables (Phase 1)
+
+### Milestone 1: Standalone Package + Edge Authorization + RecordTime Ordering
+- **Estimated Delivery:** Week 4
+- **Focus:** Extract the core ingester from Qasara's commercial gateway into a self-contained MIT-licensed package; build per-subscriber party authorization at the streaming edge; replace the buffer-based replay with RecordTime ordering and time-based resume.
+- **Deliverables / Value Metrics:**
+  - npm package decoupled from Qasara's auth, API key, and database systems; configuration via environment variables only (no Prisma, no external DB)
+  - Multi-auth support: sandbox HMAC, DevNet OAuth2 ClientCredentials, configurable per deployment
+  - **Per-subscriber party authorization** — a client may only stream parties its credential is authorized for, enforced at subscribe time; re-establishes the node's per-party visibility boundary at the streaming edge
```
hrischuk-da, line 269: Please explain the mechanism for "Per-subscriber party authorization"?

proposals/canton-event-stream.md
```diff
@@ -0,0 +1,387 @@
...
+
+### Milestone 2 — General-purpose and adoptable
+- **A new event type is added via configuration alone** (a template/interface not in the default profile is subscribed and classified) without code changes — demonstrating the subscription surface is general-purpose, not token-specific.
+- An external tester with no prior Canton experience reaches a working event stream by following only the quickstart guide, **unassisted**. *(Design target: ~10 minutes of hands-on time once base images are pulled; the gate is unassisted success, not a stopwatch.)*
+- The documented API behaves as specified: the published AsyncAPI spec renders in standard tooling, and the integration suite — WS/SSE lifecycle, stream-token single-use, RecordTime resume, invalid-token and unauthorized-party rejection, multi-party routing — passes in CI as the evidence.
+
+### Milestone 3 — It runs in production, is community-maintainable, and proposes a standard
+- ≥1 production-track deployment running alongside a validator — **Qasara's own MainNet validator as the reference deployment** — verifiable live by the committee.
```
hrischuk-da: This doesn't sufficiently demonstrate value to the community.  I would think that this solution needs to be adopted by at least 3 external community members and deployed to MainNet. 
This will increase your delivery time but it will demonstrate value.

proposals/canton-event-stream.md
```diff
@@ -0,0 +1,387 @@
...
+---
+
+## Funding
+
+**Total Phase 1 Funding Request:** 300,000 CC
+
+### Payment Breakdown by Milestone (Phase 1)
+- Milestone 1 (Standalone Package + Edge Authorization + RecordTime Ordering): **120,000 CC** upon committee acceptance
+- Milestone 2 (Configurable Subscriptions + Integration Tests + AsyncAPI Spec): **100,000 CC** upon committee acceptance
+- Milestone 3 (Production Hardening + CIP Draft + Community Onboarding): **80,000 CC** upon final release and acceptance
```
hrischuk-da: The last milestone needs to be at least 50% of the funding because it is when the ecosystem value can be best demonstrated.

proposals/canton-event-stream.md
anand-qasara: Updated the figure to show the subscribe/unsubscribe back-channel: a WebSocket client picks its parties and can change that set live (subscribe [alice], later add [bob], drop [alice]). SSE clients set their parties at connect, since SSE is one-way. Per-party authorization runs on each subscribe.

proposals/canton-event-stream.md
anand-qasara: Yes: with a party on multiple participants, each participant emits the same update, so Phase 2 dedups on the participant-independent identity. The same update carries the same updateId and RecordTime on every participant, so the ingester keeps a seen-set of updateIds within the RecordTime window and drops repeats. Updated in the phasing section

proposals/canton-event-stream.md
anand-qasara: Added in the document. Qasara's commercial offering is a hosted infrastructure service that runs this same open-source code: the grant funds the OSS package, and the hosted product consumes it as-is rather than maintaining a private fork,

proposals/canton-event-stream.md
anand-qasara: Added Token Standard V2 (CIP-0112) to the default profile alongside V1. V2 is especially relevant here because it ships a dedicated transfer-events surface (holdings-change events), so the ingester can read transfer events directly rather than inferring them from exercised choices.

proposals/canton-event-stream.md
anand-qasara: You are right, and it was not buried anywhere, it was missing. I added a payload block that carries the Token Standard semantic fields a consumer needs: instrument, amount, sender, receiver, holding. It is populated from CIP-0056 (V1) holdings/transfer-instruction data or CIP-0112 (V2) holdings-change events, so a dApp reads "Alice sent Bob 100 Amulet" directly rather than decoding the contract. The data block keeps the raw ledger metadata for anyone who needs to go deeper, and non-token event types carry their own typed payload under the same envelope.

proposals/canton-event-stream.md
anand-qasara: A profile is the configurable unit, and it has three parts: the subscribe set (interfaces/templates to pull), the classification mapping (raw event to named type), and the payload mapping (token fields to event.payload). I added a concrete default profile showing all three, covering CIP-0056 V1 (Holding, TransferInstruction) and CIP-0112 V2 (holdings-change transfer-events). Operators can extend it or ship additional profiles for non-token domains; the token-standard profile is default.

proposals/canton-event-stream.md
anand-qasara: Added an end-to-end block diagram. It shows the full path: Canton participant node, then the Ledger Ingester (one upstream subscription, classify to named events, RecordTime ordering, Phase 2 multi-host dedup), then Redis (per-party channels and a short replay buffer), then the Event Stream Server (token auth, per-party authorization, WS and SSE, RecordTime resume), then application clients, which never touch the node and pick their parties over the connection. The operator profile config feeds the ingester, and the whole service is one Docker image with Redis as its only dependency.

proposals/canton-event-stream.md
anand-qasara: I would not bet the funding on an ecosystem-adoption claim. I have reworded it to what the grant actually delivers: a team can get classified, party-authorized events by running one container, instead of building that plumbing themselves. Whether it becomes the default across the ecosystem is adoption, which I moved out of the outcome statement.

proposals/canton-event-stream.md
anand-qasara: Added Authorization model section in the document  -the server re-creates the node's per-party visibility boundary at the streaming edge using short-lived, party-scoped tokens:
Who issues tokens. The operator running the service, or their existing identity system: whoever already knows which user maps to which Canton party (for example, an exchange that knows "this logged-in user = party alice"). The server does not decide party ownership; it only accepts tokens signed by an issuer it is configured to trust (a shared key or a JWKS URL). This is the same trust model the participant already uses for actAs claims in an IdP-signed JWT, applied one hop further out.
Two deployment shapes. (a) the service authenticates the client, looks up its parties from an operator-supplied mapping, and issues the token itself; or (b) the operator's identity system issues the token and the service only verifies it. Making that verify step pluggable, so any operator can use their own identity system, is the net-new open-source work. (Qasara's commercial gateway uses shape (a) today, tied to its API-key system.)
Token contents. An explicit list of the party IDs the holder may stream, plus standard iss/aud/exp and a single-use id. The allowed-party list is written into the signed token by the issuer; the server never infers it. Tokens are short-lived (around 60 seconds) and single-use.
Subscribe-time check. Before opening any channel, the server (1) verifies the token (signature, audience, not expired, not already used), (2) reads the allowed-party list, and (3) confirms every requested party is on that list. If any requested party is not, the subscribe is rejected and no channel opens. This is exactly what the Milestone 1 acceptance test demonstrates: request a party the token does not include, and the subscribe is rejected.
Optional node-side check. Configurably, the service can also confirm with the participant that the token's issuer actually holds read rights for the party, so even a mis-issued token cannot grant more than the node itself would allow.

proposals/canton-event-stream.md
anand-qasara: I have made external MainNet adoption the primary Milestone 3 gate: at least 3 independent community members deploying the service alongside their own validator on MainNet, evidenced by their live deployment. Qasara's own gateway and validator now serve only as the reference integration and deployment, not as the value proof. This does extend the timeline as you expected: the build stays about 12 weeks, and the adoption window runs up to 6 months.

proposals/canton-event-stream.md
anand-qasara: Done. Milestone 3 is now 50% of the grant (150,000 of 300,000), released as a single payment when the full M3 criteria are met, including the external MainNet adoption above.

sunstrike228:
Read it end to end. The fanout argument is what convinced me. The moment you put a tier in front of the participant, the node's per-party visibility drops out of the downstream path, and generic pub/sub can't restore it because it has no idea what a Canton party is. Better solved once than separately by five teams.

You also state where the fan-in benefit doesn't apply instead of skipping past it, which made the rest of the numbers easier to trust.

Complementary to us on the mobile side. A phone can't hold a ledger token or keep a participant connection open, so a party-scoped edge that classifies events is what a mobile client needs. We would use this.

Two things I'd want pinned down as a consumer. Tokens are around 60 seconds and single use, but a WebSocket stays open for hours, so what happens on the live channel when a token expires or an entitlement is revoked? And RecordTime isn't unique per event, so "everything since T" needs a tiebreaker plus an inclusive or exclusive rule, or people get duplicates or a gap across reconnects.

anand-qasara:
@sunstrike228
Thanks, and you're right: a phone can't hold a ledger token or keep a live connection open, so it needs something in front that does that and only ever shows each user their own events. That's exactly what CES will handle.

On your two points:

The 60s token authorizes the handshake, not the whole session. The approach we'll pin down: give the session a bounded life with silent in-band re-auth, so a normal expiry stays invisible to the user, and handle revocation actively, so when a key or entitlement is revoked the edge closes the matching streams within seconds with a distinct close code, rather than letting a revoked client ride an open socket.

You're right that RecordTime isn't unique on its own, so we won't key resume on it alone. The plan is a composite cursor, (synchronizerId, recordTime, updateId, nodeIndex), ordered per synchronizer, resumed exclusive of the last event you processed: recordTime picks the instant, updateId the transaction in it, nodeIndex the event in the transaction. That removes the tie, so no duplicates and no gap across reconnects. We're moving to RecordTime as the key on purpose, since it's participant-independent, so a client that fails over to another host still resumes at the right spot.

proposals/canton-event-stream.md
```diff
@@ -0,0 +1,464 @@
...
+- *Two deployment shapes.* (a) the service authenticates the client, looks up its parties from an operator-supplied mapping, and issues the token itself; or (b) the operator's identity system issues the token and the service only verifies it. Making that verify step pluggable, so any operator can use their own identity system, is the net-new open-source work. (Qasara's commercial gateway uses shape (a) today, tied to its API-key system.)
+- *Token contents.* An explicit list of the party IDs the holder may stream, plus standard `iss`/`aud`/`exp` and a single-use id. The allowed-party list is written into the signed token by the issuer; the server never infers it. Tokens are short-lived (around 60 seconds) and single-use.
```
mziolekda: This can be hard to implement in the IAM systems. The mapping from user to parties is often only stored in the participant's user database and the token issuer does not have that information.

anand-qasara:
@mziolekda 

You're right, most IAMs only know people, not Canton party IDs. Making the token carry them means every operator has to sync that mapping in and keep it there. So the token just says who the user is and we look up the parties. When each user has their own ledger user we read their grants straight from the participant, so there's nothing to sync. That doesn't cover an operator running one service account across many customer parties, where the participant can't tell those users apart. So we'll make party lookup pluggable: the participant's user rights, an operator-supplied mapping, or a list in the token for issuers that already have it.

mjuchli-da:
I think this proposal addresses a gap in the Canton ecosystem, and I can see the need for an event stream service that is optimized for end-user applications (dApps). To me, the primary consumers of such a service are dApp developers, and as such I review it from a dApp architecture point of view.

Some things I particularly like:
* Keeping this as a separate Node.js component rather than embedding it into Canton (although I would consider merging once this is mature)
* Reusing/relying on the Ledger API authorization model (but I let @mziolekda and others point out the particular concerns)
* Building it using the Wallet SDK, which keeps the developer experience aligned with the rest of the ecosystem and demonstrates the intended approach for ecosystem builders.

One aspect I think is currently missing, however, is how this integrates with the dApp architecture. This is probably the most important architectural question from my perspective.

The dApp API is only aware of the Ledger API endpoint. At the same time, exposing the Ledger API is optional; some wallet providers intentionally keep it private and instead require dApps to proxy ledgerApi requests via the dApp API. I suspect the same considerations will apply to a Canton Event Stream service.

That raises a few questions:
* How does a dApp discover the Canton Event Stream endpoint (host/port) of the connected validator?
* Does this require an amendment to CIP-103 so that wallet providers can advertise the availability and location of a Canton Event Stream service (host/port)?
* Alternatively, should wallet providers be responsible for optionally proxying the event stream through the dApp API (e.g. `dAppApi.on('CES:transfer.pending')`) so that dApps never need to know where the underlying service is hosted?

If the latter approach is preferred, I think it would also be important to consider the different wallet deployment models (browser extension, desktop, remote wallet, etc.), as they each have different networking assumptions and trust boundaries.

Overall, I like the direction of the proposal. I think clarifying how a dApp discovers and consumes the event stream within the existing wallet architecture would make it much easier to see how this fits into the broader ecosystem.

hrischuk-da:
Our engineering team has provided the following feedback which requires a response, as well as Marc's excellent review above.

- Why is it simpler to re-implement authorization at the edge using JWTs instead of permissioning users on the LAPI?
- For supporting sub-party fan-out, they'd need to re-implement the privacy projection logic for stream events that we already have in the LAPI.  There is significant complexity with re-implementing home-grown party projections.  Can you provide details about this in the proposal?
- What is the backwards compatibility guarantee / compatibility model?

proposals/canton-event-stream.md
```diff
@@ -0,0 +1,464 @@
...
+2. Raw DAML contract events are classified into human-readable typed events (`transfer.pending`, `transfer.executed`, `transfer.accepted`, `transfer.rejected`, `transfer.withdrawn`, `contract.created`, `contract.archived`) via the configured mapping.
+3. Events are routed by extracting `signatories`, `observers`, `witnessParties`, and `actingParties` across event types, then fanned out to subscribers via per-party Redis Pub/Sub channels. Subscribers are authorized per party at subscribe time — a client may only stream parties its credential is entitled to, re-establishing at the edge the per-party visibility boundary the participant enforces natively (and that is otherwise lost once clients are decoupled from the node for fanout). The token-based mechanism is detailed under **Authorization model** below.
```
mziolekda: Which stream types do you intend to distribute: ledger effects or ACS-delta? Please note that the event set returned to the requester is different in these two cases. Ledger-effects stream is richer and contains all events present in ACS-delta plus more. However, reconstructing ACS-delta cannot be done as a stateless pass-through stage on the ledger-effects stream.

anand-qasara:
> Our engineering team has provided the following feedback which requires a response, as well as Marc's excellent review above.
> 
> * Why is it simpler to re-implement authorization at the edge using JWTs instead of permissioning users on the LAPI?
> * For supporting sub-party fan-out, they'd need to re-implement the privacy projection logic for stream events that we already have in the LAPI.  There is significant complexity with re-implementing home-grown party projections.  Can you provide details about this in the proposal?
> * What is the backwards compatibility guarantee / compatibility model?

@hrischuk-da 

Thanks, and we'll provide detail in the proposal.

On authorization: this isn't a replacement for LAPI permissioning, it's the same permissioning enforced one hop out. The default resolver reads the user's CanReadAs and CanActAs grants from the participant's own user management, so the party set is the node's answer rather than ours. The edge check exists because any fanout tier removes the node's enforcement from the downstream path, so whatever sits there has to re-check. The alternative is every end-user session holding its own LAPI connection and a token with real ledger authority, which is what the tier is there to avoid. We also see operators running a single shared service account rather than a ledger user per person, and there the LAPI has no per-user grants at all, so the operator's own mapping is the only source; we support both.

On projection: we don't re-implement it, we consume it. The ingester subscribes with a multi-party filter, the participant computes visibility per event, and that arrives as witnessParties on every event we handle, Created, Exercised and Archived alike. Routing is a set intersection of the event's witnessParties with the subscriber's authorized parties. No stakeholder or divulgence logic is recomputed anywhere. We also don't project below party granularity: users mapped onto one shared party all see that party's stream, because the party is Canton's unit of privacy and we keep it that way. 

On compatibility, three surfaces. The client-facing event schema is versioned and additive within a major, with the AsyncAPI spec as the contract and the Phase-1 CIP draft as the change process. Upstream we consume JSON Ledger API v2 and publish a tested compatibility matrix, Canton 3.5+ and Splice 0.5.x/0.6.x, with a 12-month maintenance commitment. Subscriptions and classification key on package-name aliases and interface IDs rather than pinned package IDs, so token-standard upgrades don't break classification, and resume cursors are participant-independent so they survive restarts and failover

anand-qasara:
Thanks @mjuchli-da , fair point, and we'll add a dApp-integration section covering it.

Providers keep the Ledger API private because it's the raw node. CES is meant to be the part you can expose: narrow, classified, authorized per party. So a provider who wants the node private gets that by running CES and exposing it, rather than hiding both. Same for hosted wallets, where one exposed CES serves all their users' sessions instead of the wallet backend building its own fanout.

Discovery is one optional field. The wallet already gives the dApp an endpoint and token through getActiveNetwork, so a provider running CES just adds the event-stream URL. No field means no CES. It's additive, so by CIP-0103's own versioning rule no amendment is needed.

The dApp needs no new credentials. It connects with the token it already has for ledger reads, and we check it against the same issuer the participant trusts. If the token names parties, we use those. If it names a ledger user, we look up that user's rights on the participant. Either way it's your authorization model reused, not a new one.

Direct connection is the default, for the reason CIP-0103 already avoids proxying reads. Your CES:transfer.pending shape stays available for fully closed setups, and since the event schema is standard, both paths deliver identical events, so dApps write one handler. For extensions, the page holds the WebSocket, not the background worker, so suspension isn't an issue.

proposals/canton-event-stream.md
anand-qasara: @mziolekda 

We consume ledger-effects upstream, and both versions of the token standard need it. In V1 the transfer outcomes are consuming choices on the same TransferInstruction, so under ACS-delta you'd be inferring the outcome from the create/archive pattern instead of reading the choice name. In V2, EventLog_HoldingsChange is nonconsuming, so it produces neither a create nor an archive and doesn't appear in ACS-delta at all.
Downstream we distribute neither shape. Subscribers get classified events like transfer.accepted, so converting between shapes never comes up.
You're right that ACS-delta isn't a stateless transform, and we're not attempting it. Consumers needing a per-party ACS view should use PQS or their own Ledger API subscription.

shaul-da:
@hrischuk-da @mziolekda we discussed this one in the voting meeting and need your input if your comments are resolved to your satisfaction and whether this is a priority. In the meantime, we're moving it to "Needs Revision" until further notice.

proposals/canton-event-stream.md
hrischuk-da: This is resolved to my satisfaction

proposals/canton-event-stream.md
hrischuk-da: This is resolved to my satisfaction

proposals/canton-event-stream.md
hrischuk-da: This is resolved to my satisfaction

proposals/canton-event-stream.md
hrischuk-da: This is resolved to my satisfaction

proposals/canton-event-stream.md
hrischuk-da: This is resolved to my satisfaction

hrischuk-da:
@shaul-da my feedback have been sufficiently resolved so this can proceed.

proposals/canton-event-stream.md
mziolekda: Thanks, that clarifies it

mziolekda:
My comments have also been answered. I am fine with proceeding

PR #109, opened 2026-03-19
Proposal: Wallet Gateway Reference Implementation

Stephan-da:
## Development Fund Proposal Submission

**Proposal file:**  
/proposals/wallet-gateway-reference-implementation.md

---

## Summary
The Wallet Gateway Reference Implementation is a secure, vendor-neutral middleware bridging user validator nodes, institutional key stores, and CIP-0103 compliant dApps. This infrastructure directly accelerates Canton Network adoption by enabling self-service institutional onboarding, allowing enterprises to securely manage real-world assets via a standardized "Clear Signing" interface without bespoke integration cycles. By decoupling signing services from application environments, the Gateway increases ecosystem velocity and eliminates redundant engineering for developers. Furthermore, it elevates zero-trust security by replacing centralized identity providers with self-signed OIDC tokens, granular access controls, and cryptographic transaction verification. To ensure guaranteed ROI, the proposal ties part of the grant request to a performance-based adoption bounty, paying out strictly upon verifiable integrations to eliminate upfront risk for the Foundation.

---

## Checklist
- [x] Proposal file added under `/proposals/`
- [x] Milestones and funding amounts defined
- [x] Acceptance criteria included
- [x] Alignment with Canton priorities described

---

## Notes for Reviewers
(Add anything the Tech & Ops Committee should pay attention to.)

PR #67, opened 2026-03-10
Proposal:  Open Sourcing Grant of a Validator Indexer (PQS)

hrischuk-da:
## Development Fund Proposal Submission

**Proposal file:**  
Link to the proposal added in this PR (e.g., /proposals/your-proposal-name.md)

---

## Summary
Briefly describe the proposal and the value it delivers to the Canton ecosystem (2–3 sentences).

---

## Checklist
- [X] Proposal file added under `/proposals/`
- [X] Milestones and funding amounts defined
- [X] Acceptance criteria included
- [X] Alignment with Canton priorities described

---

## Notes for Reviewers
(Add anything the Tech & Ops Committee should pay attention to.)

PR #318, opened 2026-05-08
Proposal: Denex Localnet

mgaare:

**Proposal file:**  
./proposals/denex-localnet.md

---

## Summary

`denex-localnet` is a Testcontainers-style SDK and CLI for running Canton Network LocalNets, collapsing what would currently require dozens of scattered config files into a single declarative YAML file. It orchestrates the full Canton stack — Super Validator, regular validators, Splice, Keycloak auth, and all associated web UIs — directly via the Docker API (no Docker Compose), with sensible defaults for ports, party hints, realm naming, and OAuth2 wiring.

It lowers the barrier to local development and integration testing: developers can spin up a fully-wired multi-validator network programmatically (via the cross-runtime SDK on Deno/Node/Bun) or interactively (via the Deno CLI), then query state, allocate parties, and provision users through clean high-level APIs. This enables programmatic and flexible integration for app developers, and makes Canton LocalNets viable as ephemeral fixtures in CI pipelines and SDK-driven test suites — a workflow that's currently very challenging with existing tooling.

---

## Checklist

---

## Notes for Reviewers

### DIFF
--- a/proposals/denex-localnet.md
+++ b/proposals/denex-localnet.md
@@ -1,0 +1,609 @@
+# Denex Canton Localnet
+
+# 1. Executive Summary
+
+We propose the continued development and open-source release of the
+**Denex Localnet**, addressing one of the higest-rated "Critical"
+needs identified in the [2026 Developer Experience and Tooling
+Survey](https://discuss.daml.com/t/canton-network-developer-experience-and-tooling-survey-analysis-2026/8412).
+
+**Denex Localnet**: A configurable local Canton network manager that replaces
+the current approach of hardcoded, multi-file configuration with a single
+declarative configuration file, programmatic API, and CLI for orchestrating,
+inspecting, and modifying running local networks.
+
+**Funding Requested:** 8,500,000 CC **Timeline:** 9 months **Team:** 4
+developers
+
+---
+
+# 2. Problem Statement
+
+41% of respondents cited environment setup and node operations as the task that
+took the longest to "get right." Developers are forced to be **infrastructure
+engineers before they can be product builders**. The existing Splice LocalNet
+and CN Quickstart approaches require navigating 80-120+ configuration files
+spread across Docker Compose fragments, environment files, HOCON configs,
+Makefiles, Gradle tasks, and Keycloak realm exports, with hardcoded party hints,
+port numbers, OAuth client IDs, and user identifiers scattered throughout.
+
+The survey's highest-urgency rating went to **Local Development
+Frameworks**, where developers specifically name-dropped Hardhat,
+Foundry, and Anchor as the standard they expect. They want composable,
+flexible, and integratable infrastructure tooling for app development,
+testing, CI/CD, and other key flows. Currently, they are "cobbling
+together" scripts.
+
+There are several different use cases and development flows that we
+are targeting:
+- Local app development, particularly with multiple participants
+- Integration and end-to-end testing
+- `testcontainers`-style test orchestration
+- CI/CD flows
+- Demos and other repeatable/controlled environments
+
+---
+
+# 3. Proposed Solution: Denex Localnet, a Configurable Local Canton Network Manager
+
+**Category:** Local Development Framework
+
+**What it does:** A tool to run, configure, and control Canton local networks
+for development and testing, with a single-file declarative configuration model
+and both CLI and programmatic APIs.
+
+**Key technical features:**
+
+- **Single-file configuration** for the entire localnet topology: number of
+  participants/validators, party and user onboarding, permissioning, and
+  network ports, all existing in one file with straightforward syntax
+- **Programmatic API** for configuring and modifying local networks
+  from application code and test suites, supporting a
+  `testcontainers`-style workflow
+- **CLI tool** for managing running localnets: start, stop, modify
+  topology, onboard parties, inspect state, inspect state and
+  metadata, and query tools to support programmatic creation of app
+  .env files
+- **Runtime state inspection:** Query a running localnet for active users,
+  ports, parties, and configuration, enabling programmatic integration without
+  hardcoding parameters
+
+**Impact:** Directly addresses the #1 survey finding. Transforms the current
+experience of navigating 80-120+ files across Splice LocalNet and CN Quickstart
+into a single, readable configuration. Developers can bring up a customized
+Canton network in minutes rather than hours, and integrate it programmatically
+into their test suites and CI/CD pipelines.
+
+## Examples
+
+Not all details have been finalized, so aspects of these examples are
+subject to change. These are intended to help illustrate the design
+thinking and capabilities of `denex-localnet`.
+
+### Configuration Files
+
+Configuration for `denex-localnet` is specified in one of two ways: in
+a single .yaml file, or via a fluent TypeScript API (examples in below
+section). Both support the full featureset.
+
+Here is an example configuration file that is intentionally more
+verbose than a realistic user's file, in order to demonstrate features.
+
+```yaml
+# Optional. Base port for allocation. SV uses ports starting here; each
+# validator gets +100 from the previous one. Default: 5000.
+#   SV:           6000-6099  (Keycloak at 6082, web UIs at 6080)
+#   alice:        6100-6199  (wallet UI at http://wallet.localhost:6180)
+#   bob:          6200-6299  (wallet UI at http://wallet.localhost:6280)
+#   admin-vldr:   6300-6399  (wallet UI at http://wallet.localhost:6380)
+basePort: 6000
+# ---------------------------------------------------------------------------
+# Keycloak OIDC backend
+# ---------------------------------------------------------------------------
+# The Keycloak admin user. Used at startup to bootstrap the realm imports
+# AND at runtime by KeycloakAdminClient for runtime user provisioning.
+# The admin/password convention is: username == password (everywhere).
+auth:
+  keycloak:
+    admin: admin
+    password: admin
+# ---------------------------------------------------------------------------
+# Validators
+# ---------------------------------------------------------------------------
+# The Super Validator (SV) is IMPLICIT — always exactly one, created
+# automatically. You only configure the *regular* validators here.
+#
+# Two forms are supported:
+#
+#   validators: 2            # shorthand — creates validator-1, validator-2
+#
+#   validators:              # detailed — custom names, parties, users
+#     - name: ...
+#
+# IMPORTANT: validator names must produce ≤11 chars after `-` → `_`
+# transformation, because Splice enforces a 30-char limit on the derived
+# participant node name (`{participantName}-validator_backend`, where
+# `participantName = name.replace(/-/g, '_')`).
+#
+# Examples that work:  alice, bob, val1, app, admin-vldr
+# Examples that fail:  alice-validator (33 chars after suffix → crashes splice)
+validators:
+  # -------------------------------------------------------------------------
+  # Validator 1: minimal — just a name
+  # -------------------------------------------------------------------------
+  # Gets a Participant + Validator App + wallet UI. No application parties
+  # or users (just the operator's own Splice-managed wallet-admin user).
+  - name: alice
+  # -------------------------------------------------------------------------
+  # Validator 2: full feature set — parties + users with all rights variants
+  # -------------------------------------------------------------------------
+  - name: bob
+    # Parties allocated on this validator's Participant at startup.
+    # Hints become part of the full party ID. Users referencing these
+    # hints (or any other hint) auto-allocate them if not pre-listed.
+    parties:
+      - hint: bob                       # minimum: just a hint
+      - hint: trader                    # explicit display name
+        displayName: Trader Joe
+      - hint: auditor
+        displayName: Auditor
+    # Users created on this validator's Participant. Each user is fully
+    # provisioned end-to-end at startup: ledger account + Keycloak realm
+    # account + Splice wallet onboarding.
+    #
+    # Username == password convention (so 'bob-trader' logs in with
+    # password 'bob-trader'). The Keycloak realm for this validator is
+    # 'Bob' (title-cased from validator name).
+    users:
+      # User with a primary party only — gets CanActAs(bob) automatically
+      - id: bob-trader
+        primaryParty: bob
+      # User with primary party AND additional per-party rights
+      - id: bob-readonly
+        primaryParty: bob
+        parties:
+          - hint: trader                # defaults to [CanActAs]
+          - hint: auditor
+            rights: [CanReadAs]         # explicit rights list
+      # User with multiple per-party rights, no primary party
+      - id: cross-reader
+        parties:
+          - hint: bob
+            rights: [CanReadAs]
+          - hint: trader
+            rights: [CanReadAs, CanExecuteAs]
+      # User referencing an UNDECLARED party — auto-allocated at startup
+      - id: dynamic-user
+        primaryParty: late-bound        # not in `parties:` above; auto-allocated
+        parties:
+          - hint: also-new              # also auto-allocated
+            rights: [CanActAs]
+      # Admin-only user — no party at all
+      - id: bob-admin
+        rights:
+          - ParticipantAdmin            # full admin access on this Participant
+      # User with multiple participant-wide rights (admin variants)
+      - id: bob-superuser
+        rights:
+          - ParticipantAdmin
+          - CanReadAsAnyParty
+          - CanExecuteAsAnyParty
+          - IdentityProviderAdmin
+  # -------------------------------------------------------------------------
+  # Validator 3: short-named, focused on package upload routing
+  # -------------------------------------------------------------------------
+  # `admin-vldr` becomes `admin_vldr` in the participant name (10 chars,
+  # leaving room for `-validator_backend` → 28 chars total ≤ 30).
+  # Realm: 'AdminVldr'.
+  - name: admin-vldr
+    parties:
+      - hint: ops
+    users:
+      - id: ops-user
+        primaryParty: ops
+        rights: [ParticipantAdmin]
+# ---------------------------------------------------------------------------
+# DAR packages to upload after startup
+# ---------------------------------------------------------------------------
+# Each package is uploaded to the listed validators. Default: all validators.
+# Paths are resolved relative to YAML file's directory.
+packages:
+  # Upload to all validators (and the SV)
+  - name: my-app
+    dar: ./dars/my-app.dar
+  # Upload to specific validators only
+  - name: bob-only-extension
+    dar: ./dars/bob-extension.dar
+    uploadTo:
+      - bob
+  # Upload to a subset
+  - name: shared-types
+    dar: ./dars/shared-types.dar
+    uploadTo:
+      - alice
+      - bob
+```
+
+---
+
+### Typescript API
+
+1. Quickstart — config from YAML
+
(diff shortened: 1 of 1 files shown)


proposals/denex-localnet.md
```diff
@@ -0,0 +1,609 @@
...
+      - hint: auditor
+        displayName: Auditor
+    # Users created on this validator's Participant. Each user is fully
+    # provisioned end-to-end at startup: ledger account + Keycloak realm
+    # account + Splice wallet onboarding.
+    #
+    # Username == password convention (so 'bob-trader' logs in with
+    # password 'bob-trader'). The Keycloak realm for this validator is
+    # 'Bob' (title-cased from validator name).
+    users:
```
da-tanabe, line 157: From an implementation standpoint, have you seen the declarative API for users/parties? (https://docs.digitalasset.com/operate/3.4/howtos/configure/general/declarative_conf.html)

It would be nice if, as part of this, changes were implemented in lower levels of the stack so that components understand declarative APIs more natively.

mgaare: I hadn't seen that before. It looks interesting.

`denex-localnet` also provisions the users in keycloak from this config, and it's network-wide rather than localized to a single participant, so a 1-1 correspondence with the dynamic participant config feature probably isn't possible. It's worth thinking about using it in the implementation, though.

vadasnorbert:
Based on our experience with Canton dev environments, a “Hardhat-like” tool could be very useful for Canton, and could significantly improve the developer experience for both new and experienced Canton builders. 

The proposal also directly addresses one of the clearest pieces of feedback from the DevEx survey, i.e. the need for better local development frameworks and simpler environment setup. In addition, features described in the proposal, such as the single-file declarative configuration, programmatic API for lifecycle, or the runtime state inspection/discovery could potentially offer capabilities beyond those of existing frameworks. Personally, I am supportive of this proposal.

proposals/denex-localnet.md
daravep: Note that mid term we wanted to decouple the user database from the validator instance in order to use it across the different node types and support direct integration with existing user management systems rather than to hand-roll our own. This is more of a heads-up, though.

proposals/denex-localnet.md
```diff
@@ -0,0 +1,609 @@
...
+# The admin/password convention is: username == password (everywhere).
+auth:
+  keycloak:
+    admin: admin
+    password: admin
+# ---------------------------------------------------------------------------
+# Validators
+# ---------------------------------------------------------------------------
+# The Super Validator (SV) is IMPLICIT — always exactly one, created
+# automatically. You only configure the *regular* validators here.
```
daravep, line 114: What exactly will constitute a validator here? Participant node & validator app?

proposals/denex-localnet.md
```diff
@@ -0,0 +1,609 @@
...
+#   validators: 2            # shorthand — creates validator-1, validator-2
+#
+#   validators:              # detailed — custom names, parties, users
+#     - name: ...
+#
+# IMPORTANT: validator names must produce ≤11 chars after `-` → `_`
+# transformation, because Splice enforces a 30-char limit on the derived
+# participant node name (`{participantName}-validator_backend`, where
+# `participantName = name.replace(/-/g, '_')`).
+#
```
daravep, line 127: Note that multi-synchronizer support is coming (and is already usable for alpha development in 3.5). Does the grant here include running multiple synchronizers as well? 

Also please note that recently PQS was approved as OSS, which means that it is likely going to become a standard piece of infrastructure for app provider backends: https://github.com/canton-foundation/canton-dev-fund/pull/67

proposals/denex-localnet.md
mgaare: Yes - participant node, validator, wallet ui, the full splice stack.

proposals/denex-localnet.md
```diff
@@ -0,0 +1,609 @@
...
+});
+
+// SV-specific user — realm name automatically resolved to 'SV' (not 'Sv')
+await net.createUser('sv-helper', 'sv', { rights: ['ParticipantAdmin'] });
+```
+
+Idempotency: re-calling with the same args is safe — each side (ledger, Keycloak, wallet) converges. Partial failures are recoverable; just call again.
+
+---
+
```
daravep, line 310: Wouldn't it make more sense to base the CLI  here on a normal Typescript Client that can also be used in applications and mostly provide a factory to automatically discover and connect to the node? This would allow devs to only learn one syntax API and avoid creating a small subset that only works in a few common cases.

proposals/denex-localnet.md
mgaare: Not at this point. I don't have a good enough sense of the use cases and shape of the requirements there to have a point of view on what/how to build.

proposals/denex-localnet.md
```diff
@@ -0,0 +1,609 @@
...
+Canton network in minutes rather than hours, and integrate it programmatically
+into their test suites and CI/CD pipelines.
+
+## Examples
+
+Not all details have been finalized, so aspects of these examples are
+subject to change. These are intended to help illustrate the design
+thinking and capabilities of `denex-localnet`.
+
+### Configuration Files
```
daravep, line 83: How are you going to support passing custom configuration arguments to the individual nodes?

proposals/denex-localnet.md
mgaare: Not sure I quite follow what you're saying here.

`denex-localnet` provides two interaction surfaces, which have essential feature parity between them:
- A TypeScript API (documented in this section) with which a user can create, configure, lifecycle, query status and metadata, and to an extent modify localnets from that user's own code
- A CLI tool that can be used to do the same things

With those tools you can both create a new localnet or interact with a running localnet - and there could be more than one running. This section is showing how the TypeScript API can get a handle on a running localnet. The handle returned from this is the same type of object you get when creating a new localnet from a config.

proposals/denex-localnet.md
mgaare: What kind of configuration do you have in mind?

proposals/denex-localnet.md
```diff
@@ -0,0 +1,609 @@
...
+  .env files
+- **Runtime state inspection:** Query a running localnet for active users,
+  ports, parties, and configuration, enabling programmatic integration without
+  hardcoding parameters
+
+**Impact:** Directly addresses the #1 survey finding. Transforms the current
+experience of navigating 80-120+ files across Splice LocalNet and CN Quickstart
+into a single, readable configuration. Developers can bring up a customized
+Canton network in minutes rather than hours, and integrate it programmatically
+into their test suites and CI/CD pipelines.
```
daravep, line 75: How are you going to solve this, actually, technically. Do you plan to map this to docker? Do you plan to integrate this tightly with the splice releases? 


How do you minimize resource usage? This has been an issue with localnet, so it's worth bringing it up here so you don't run into the same problem. Right now, a full localnet creates too many resource hungry JVM processes. 

canton.jar itself does support running many nodes in a single jvm process (and in theory wouldn't be hard to support dynamically adding new nodes to a running process), but the validator app is a separate jar.

mgaare: Yes, the working beta version of `denex-localnet` uses the Docker socket directly to create and manage the containers, using configs that it generates from the single `denex-localnet` config.

I think there is an effective minimum resources needed to run a localnet. We're running a single canton container, and single validator app container, each hosting all of the nodes. All told, between canton, all the splice apps, postgres, nginx, keycloak, right now a `denex-localnet` with 2 validators configured stands up 10 docker containers consuming around 5 GB of memory.

proposals/denex-localnet.md
daravep: New features and capabilities are often introduced via feature flags so people can try them out (either via canton.parameters.xyz or canton.<nodetype>.node.parameters). Also, the resource consumption of the nodes can be tweaked there etc. Or the protocol version to be used etc: https://docs.canton.network/sdks-tools/development-tools/sandbox#dev-protocol

proposals/denex-localnet.md
daravep: Yes. I do see that one part of the Typescript API is interacting with the node lifecycle. However, the second part is ultimately interacting with the node state via Ledger API. 

In your other proposal, you are also proposing a new Typescript Ledger interface: https://github.com/canton-foundation/canton-dev-fund/pull/238/changes#diff-4f2822a1b669831f32b246ebe4d0c799c9407081c67470b41c55ec781b794b5cR17-R18

So my question is how much overlap there is and how these two will be used in conjunction.

proposals/denex-localnet.md
```diff
@@ -0,0 +1,609 @@
...
+
+| Milestone; Deadline      | Deliverable           | Requested Funding (CC) | Description                                          |
+|--------------------------|-----------------------|------------------------|------------------------------------------------------|
+| 2.1; Approval + 6 months | localnet distribution | 1,000,000              | Self-contained single file distribution for cli tool |
+| 2.2; Approval + 6 months | Documentation         | 250,000                | Unified documentation site                           |
+| 
+
+**Phase 2 Funding Requested: 1,250,000 CC**
+
+## Phase 3: Hardening
```
daravep, line 511: I would suggest that you add another work item to continuously align localnet with the splice release cadence. The artifacts you are coding against are not static (e.g. multi-sync ...). I would recommend putting an initial but low estimate in there, but refer to a possible extension grant if the effort exceeds the budget.

proposals/denex-localnet.md
hrischuk-da: I have some additional questions:

1. Does that configuration include the synchronizer containers (e..g, mediator, sequencer)?
2. How do you plan on keeping your configurations consistent with DevNet/TestNet/MainNet which may each have different config values?
3. Regarding memory, I think you need to have a requirement that the memory footprint is equal to or less than the current LocalNet for the same app.  Can you add this?
4. If a dApp needs to include other services (e.g., LAPI client, Nginx running a UI), how easily will this framework incorporate this?  For example, the dApp user needs to run a LAPI client, how will it know the port number for the LAPI for the target validator?  How will the Docker vpn accommodate these non-LocalNet images?

proposals/denex-localnet.md
hrischuk-da: Supporting PQS is a requirement.  A large percentage of dApps use PQS.

hrischuk-da: In the short term, supporting a Wallet Gateway will also be needed.  See [Proposal: Wallet Gateway Reference Implementation #109](https://github.com/canton-foundation/canton-dev-fund/pull/109).

I would expect the pattern of running sidecar, Docker images to a validator will be a recurring pattern.

proposals/denex-localnet.md
```diff
@@ -0,0 +1,609 @@
...
+  # Realm: 'AdminVldr'.
+  - name: admin-vldr
+    parties:
+      - hint: ops
+    users:
+      - id: ops-user
+        primaryParty: ops
+        rights: [ParticipantAdmin]
+# ---------------------------------------------------------------------------
+# DAR packages to upload after startup
```
hrischuk-da, line 206: A feature that we've found extremely helpful is to automatically forward log files to a single directory so that a tool like "lnav" can be used.  Can you include this?

proposals/denex-localnet.md
```diff
@@ -0,0 +1,609 @@
...
+const aliceParties = await net.getParties('validator-1');
+
+const users = await net.getUsers('validator-1');
+const usersWithRights = await net.getUsersWithRights('validator-1');
+for (const u of usersWithRights) {
+  console.log(`${u.id}: ${u.rights.length} rights`);
+}
+
+// One snapshot of everything
+const snapshot = await net.getSnapshot();
```
hrischuk-da, line 348: Please explain what `getSnapthot` does and what its boundaries are (e.g., include PQS or not)?

proposals/denex-localnet.md
```diff
@@ -0,0 +1,609 @@
...
+const packages = await net.getPackages();
+```
+
+---
+
+6. Mutations — parties and DARs
+
+```ts
+// Allocate an additional party (createUser does this automatically for referenced hints)
+const party = await net.allocateParty('alice', 'validator-1', 'Alice');
```
hrischuk-da, line 362: What kind of party is this: internal or external?  External parties are the focus because they support wallets.

Can you reuse the TypeScript signatures of the Wallet SDK and dApp SDK so that users don't need to learn several libraries that do the same thing?  For example, see https://docs.canton.network/integrations/wallet/guidance#create-an-external-party-wallet.

proposals/denex-localnet.md
```diff
@@ -0,0 +1,609 @@
...
+  env           - Show environment info for the LocalNet
+  credentials   - Show login credentials for web UIs
+  instances     - List running LocalNet instances
+  entitlements  - List users with their rights on the LocalNet
+  discovery     - Discovery server for managing multiple LocalNet instances
+```
+
+---
+
+## 4. Comparison with Existing Solutions: Denex Localnet vs. CN Quickstart / Splice LocalNet
```
hrischuk-da, line 457: Yes, you have highlighted gaps.  Will you be replacing LocalNet in Splice and CN QuickStart with denex-localnet?  If not, why?  I think that would make sense.  After all, how many types of localnet do we need?

proposals/denex-localnet.md
mgaare: There's some limited overlap, in particular for the mutations that localnet can do. Right now the beta version does not use the interface from our SDK for the things it needs to do via the ledger, mostly for some logistical/release coordination reasons. I would guess that once there's a finalized release of our SDK and npm distribution it's likely that'll get integrated into `localnet`, but don't know that for sure just yet. This is more in the realm of implementation detail.

Here is the vision for usage. `denex-localnet` is meant to stand up the Canton infrastructure-layer pieces that the application layer needs, providing a flexible way for users, applications or app test/run harnesses to configure and run that infrastructure, and supplying the information those applications would need to configure and bootstrap themselves (parties, users, network endpoints) - the sorts of things that often get plugged into say an app .env file. Beyond this initialization, consumers should generally handle all subsequent ledger interactions and state mutations on their own - perhaps by using the tools in our separate SDK, but that kind of decision is app- and usecase- specific and outside of `denex-localnet`'s bounds.

proposals/denex-localnet.md
mgaare: Those are interesting ideas and I'll have to give that some thought.

proposals/denex-localnet.md
mgaare: 1. Yes, it includes the synchronizer and the full Splice single-node Super Validator setup.
2. I'm not sure which configurations you mean. Talking about, eg, the Amulet config and associated Splice configs that are part of network governance, or some other config?
3. So I think this might be somewhat difficult, as splice and cn-quickstart localnet have a fixed number of participants, whereas `denex-localnet` has a configurable number of participants. Additionally, some of the other configuration options requested by @daravep in a separate comment, if really supported properly, would probably need to contemplate different low-level configuration options per-participant, which could in turn require `denex-localnet` to run those participants in a less memory-efficient way (separate JVM/containers per participant, potentially, rather than single multi-hosting).
4. `denex-localnet` is designed to support flexible integration with apps. It binds host ports in a configurable range for the ledger API, admin API, and the various Splice UIs/apps. The TypeScript API and the CLI are designed to support both manual and programmatic querying for what those endpoints are against a running localnet, and integrating flows to generate app .env files that apps can hook into. It allows developers to choose whatever mechanism they want to run their own app level components locally, and multiple options for how they want to bootstrap and integrate them with the underlying Canton layer.

proposals/denex-localnet.md
mgaare: So I had intended here to convey that idea with the "regular updates to incorporate Splice releases". Are you suggesting to go further and specify a release cadence that explicitly matches Splice's (currently weekly)?

proposals/denex-localnet.md
mgaare: I'll have to give that some thought, on the right way to handle this.

proposals/denex-localnet.md
mgaare: This returns an object that shows the full configuration/metadata from the localnet: parties, users, endpoints, participants, DSO, uploaded Dars, etc. It's not talking about database or runtime state. Perhaps not the best named method.

proposals/denex-localnet.md
mgaare: These are internal parties.

I will say that I have mixed feelings about including support for Mutations at all in the API. As you point out, that really starts to get into the domain of application logic rather than the infrastructure layer. There is a good chance of dropping this feature entirely, and limiting `denex-localnet` to only creating users and parties from config as part of the localnet initialization process.

proposals/denex-localnet.md
mgaare: Good tools are opinionated. While it's not for me to say whether `denex-localnet` is a good tool, it is an opinionated one. Although I have made an effort to focus the design such that the "blast radius" of its opinions are kept smaller, and as much useful choice is left up to the user, nonetheless people are free to accept them or not. Given that, I was not planning to try to promote this into Splice or QuickStart.

proposals/denex-localnet.md
mgaare: PQS, Wallet Gateway, and any other tools that a user would like to use with their app are supported the same way that apps themselves are supported: `denex-localnet` provides infrastructure-level configuration and primitives, and a rich API/CLI surface with which users can configure and run those higher-level components in whatever way they want. Whether to run PQS, Wallet Gateway, or any of a slew of other current or future Canton tools and applications, and how they are configured, is a matter of user choice.

I do think it would be a good idea to include some rich recipes in the documentation for how to configure and run apps and app tools on a `denex-localnet`.

proposals/denex-localnet.md
daravep: Apologies, looks like I've overlooked this. In that sense it seems all good.

hythloda (approved):

pedrodneves:
## Phase 1: Open-Source Release

| Milestone; Deadline      | Deliverable              | Requested Funding (CC) | Description                                                 |
|--------------------------|--------------------------|------------------------|-------------------------------------------------------------|
| 1.1; Approval + 3 months | Localnet initial release | 1,000,000              | Command line utility and API to manage local Canton network |

**Phase 1 Funding Requested: 1,000,000 CC**

pedrodneves:
## Phase 2: Ecosystem Integration

| Milestone; Deadline      | Deliverable           | Requested Funding (CC) | Description                                          |
|--------------------------|-----------------------|------------------------|------------------------------------------------------|
| 2.1; Approval + 6 months | localnet distribution | 1,000,000              | Self-contained single file distribution for cli tool |
| 2.2; Approval + 6 months | Documentation         | 250,000                | Unified documentation site                           |
| 

**Phase 2 Funding Requested: 1,250,000 CC**

pedrodneves:
## Phase 3: Hardening

| Milestone; Deadline      | Deliverable                 | Requested Funding (CC) | Description                                                                                                                                                                     |
|--------------------------|-----------------------------|------------------------|---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
| 3.1; Approval + 9 months | Localnet hardening releases | 1,000,000              | Harden configuration validation, regular updates to incorporate Splice releases and a selection mechanism; improve CI/CD integration based on community experience and feedback |

**Phase 3 Funding Requested: 1,000,000 CC**

PR #400, opened 2026-06-01
Add Sigli.ai DAWE Proposal

ayanuali:
This proposal seeks funding to build and open-source the DAML Automated Workflow Evaluator (DAWE).

## Checklist

SIG: daml-tooling

Champion: 5NORTH

### DIFF
--- a/proposals/sigli-ai
+++ b/proposals/sigli-ai
@@ -1,0 +1,309 @@
+# Sigli.ai: DAML Automated Workflow Evaluator (DAWE) for Canton
+
+## Abstract
+Sigli.ai is building the on-chain neobank for autonomous AI agents on Canton: programmable identities, wallets, compliance guardrails, and financial execution infrastructure for machines. For the identity, wallet, and payment layers, we are building on existing infrastructure, including Tenzro’s agent identity, MPC wallet, and settlement components, rather than rebuilding primitives from scratch.
+
+What does not yet exist in the Canton ecosystem is a standardised framework for automatically testing and verifying the security of DAML smart contracts before they reach mainnet. Every team building complex DAML applications today writes bespoke test scaffolding by hand, with no shared standard for what a security-verified DAML workflow looks like. For an enterprise deploying financially active smart contracts on an institutional ledger, this is not a minor inconvenience. It is a blocker.
+
+This proposal seeks ~337k CC funding to build and open-source the DAML Automated Workflow Evaluator (DAWE): a scenario-based automated security and QA testing framework for DAML applications on Canton. The grant deliverable is the test harness itself, along with a set of realistic financial workflow examples against which it runs. We are proposing to open-source the testing framework because every serious Canton developer needs it.
+
+## Background
+AI agents are financially homeless. They can execute code, manage complex workflows, and interact with services autonomously, but they cannot hold money, receive payment, or transact without a human approving every step. That is not automation.
+
+Sigli.ai is building the neobank for AI agents on Canton: identity, wallets, payments, compliance, and yield, built for machines rather than humans. The target customer is an enterprise: a bank, asset manager, treasury desk, or financial infrastructure provider that wants to deploy financially autonomous agents in a way that satisfies governance committees, internal security reviews, and regulatory requirements.
+
+For the underlying infrastructure, we are building on what already exists:
+- **Agent identity and wallets**: Tenzro provides W3C DID-based agent identity with auto-provisioned MPC wallets, threshold-signed by network validators and portable across chains. We plan to use these components for Sigli’s identity and custody layer rather than rebuilding them on Canton from scratch.
+- **Agentic payments**: Tenzro natively supports x402, MPP, Visa TAP, and Mastercard Agent Pay. Where applicable, Sigli will integrate Tenzro’s payment stack for agent-to-agent and agent-to-service payment flows.
+- **Financial settlement**: Tenzro’s financial settlement infrastructure supports atomic multi-chain settlement and runs DAML smart contracts with unified settlement alongside EVM and SVM. We plan to use this as the settlement substrate for Sigli’s neobank.
+- **DAML development environment**: Tenzro’s DAML Studio provides an AI-assisted Canton development environment for building, testing, and deploying DAML contracts. We are building on top of this toolchain.
+
+Tenzro is our current preferred provider for these components. Where any component is unavailable or does not meet requirements, we will substitute equivalent infrastructure, whether from Canton’s own emerging standards or from alternative providers. The DAWE grant deliverable itself has no dependency on Tenzro or any specific third-party infrastructure.
+
+The gap we are addressing is not in any of these layers. It is a cross-cutting need that every team building complex DAML applications on Canton faces: a standard framework for automatically testing and verifying the security of DAML smart contracts before they reach mainnet. That framework does not exist. We are building it, for ourselves and for the ecosystem.
+
+## The Problem
+Canton and DAML give developers strong formal guarantees at the contract level. What they do not provide is a standard framework for:
+- Automatically executing scenario-based tests across all contract state transitions, not just happy-path cases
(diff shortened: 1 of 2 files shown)


ayanuali (commented):
Updated champion and proposal seeks

ayanuali:
@hrischuk-da  @monsieurleberre Hi! We believe that DAWE’s purpose is most aligned with your DAML Language and Dev Tooling SIG. Request you to please take a look at our proposal and give your feedback. Happy to share any additional details that you may need.

ayanuali:
@heslin-zenith @monsieurleberre following up on this one. Any initial thoughts on the DAWE proposal? Happy to answer questions or tighten the scope if that helps.

ayanuali:
@nikos-metrika @leonidr-c7 @pdomenighetti @Ian-avro bumping the DAWE proposal above. Would appreciate any feedback, or a steer toward the right group if this SIG isn't the fit. Happy to walk through it on a call if that's easier.

monsieurleberre:
Hi @ayanuali, sorry for the slow reply.

Two questions, and I may have misread things.

First, what's the contribution exactly? I get two answers from the proposal. In places it says the tooling doesn't exist, but submitMustFail already tests that the wrong party can't exercise a choice, and Splice's test package adds expectErrorId and expectFailureStatus so you can check it failed for the right reason. Elsewhere it reads as a convention rather than a tool: a standard shape for a security suite, three worked financial examples, less boilerplate for everyone. I find the second much more convincing. Is that the pitch?

Related, you describe rejection as verified "not in off-chain assertion code that can be bypassed". I don't follow that one. With submitMustFail the ledger does the rejecting and the script only watches. What's bypassable there?

Second, the harness. WorkflowScenario is a template, so RunScenario runs every step inside a single transaction, and those commit all-or-nothing. If step 3 is rejected, doesn't the whole thing abort and take ScenarioResult with it? The "FAIL: " <> diagnostics results branch is the part I can't place, since an aborted transaction has nothing left to report with. And if executeStep avoids provoking a real rejection so the results do survive, then it's AC 3 that gives instead. Is the runner meant to be a Daml Script? Catch would be the other way out, but exceptions are deprecated now.

Please help me understand what I am missing, thanks!

ayanuali:
@monsieurleberre thank you — you have not misread it. On both points, the proposal needs correction and narrowing.

1. The contribution
Yes: the second interpretation is the intended pitch. DAWE is not a replacement for submitMustFail, nor a claim that Daml lacks negative-testing primitives. We should have said this clearly.

The contribution is a maintained, reusable security-scenario convention and reference suite for financial Daml workflows, built on the existing testing facilities. Concretely:
- a common scenario structure;
-  reusable multi-party, policy, time, and lifecycle fixtures;
- three worked financial suites;
- documented application bindings;
- CI-friendly reporting.

The value is in making the adversarial cases, setup, assertions, and evidence output reusable across teams — not in inventing a new mechanism for asserting a rejection.

Where an exact reason matters, DAWE will compose the existing primitives. In a Splice-based test, that can mean submitMustFail' with expectErrorId or expectFailureStatus; the public Splice test utilities use precisely this pattern.[1] The project will document which checks rely on a stable error ID/status and which only assert rejection plus post-state invariants.

The phrase "not in off-chain assertion code that can be bypassed" was badly phrased. In the submitMustFail case, the ledger is indeed performing the rejection and the script is observing/asserting it. The intended distinction was only between:
- an invalid command rejected by the contract/ledger, and
- an application client deciding not to submit the command because of an optional pre-check.

A test suite can naturally be skipped or changed, but that is a testing-governance issue, not a ledger-enforcement issue.

We will replace the language with:

`The scenario attempts an invalid exercise. It passes only if the ledger rejects that submission, and then verifies that the relevant ledger state did not advance. This verifies contract-level enforcement rather than relying solely on an application-side pre-check.`

We will also narrow the proposed PolicyConfig layer. It should be a test-fixture and application-binding interface, not a claim that different identity or guardrail systems can be normalized into one universal enforcement semantics. An integration must state the specific policy action and expected rejection it tests.

2. The harness and transaction model
You are also correct about WorkflowScenario / RunScenario. The sketch is not valid for a scenario that triggers a real rejection:
- Since the choice runs as one transaction, an invalid executeStep aborts that transaction, and no ScenarioResult can be returned or committed.
- Conversely, if it avoids the real rejection in order to return a ScenarioResult, it would not satisfy the acceptance criterion.

The "FAIL: " <> diagnostics branch therefore has no place in that on-ledger choice.

We will remove that template/choice sketch. The runner is a Daml Script test runner with an external report adapter, not an on-ledger WorkflowScenario contract. This does not use try/catch.

Each named scenario will be a top-level Script () test, run independently. A multi-step workflow scenario makes separate submissions for its valid state transitions; the adversarial action is then a separate expected-failure submission. For a policy violation, the test will:

1. Set up the parties, policy fixture, and workflow state.
2. Submit the valid preceding workflow steps.
3. Attempt the invalid AdvanceStep using submitMustFail, or the Splice checked variant when the test must distinguish an expected stable error ID/status from another failure.
4. Query/assert that the workflow has not advanced and that no prohibited effect occurred.

If the invalid submission succeeds, or if a checked test fails for the wrong reason, that scenario test fails. The test framework reports that outcome. There is no attempt to persist a result from the rejected transaction.

This matches the intended Daml testing model: submitMustFail exists specifically to test an expected rejected submission while allowing the script to continue, and Daml Script can also be run against a local Canton Sandbox through dpm script for the integration target.

The architecture should therefore be:

image

We will revise the affected acceptance criterion accordingly. For example:

`At least one policy-violation Daml Script scenario, executed against Canton Sandbox, attempts an invalid command that the ledger rejects. Where the applicable integration exposes a stable error identifier or status, the scenario asserts that expected reason. The scenario then verifies the workflow state did not advance, and the external runner emits a machine-readable result.
`
The unauthorized-controller and replay tests will follow the same pattern.

This is a meaningful design correction, and we appreciate you calling it out. We will update the proposal so that the scope and execution model are unambiguous.

Unless you have additional concerns with **this revised test-execution model, we will make the corresponding edits and proceed on that basis. We'd welcome any further feedback before doing so.**

ayanuali:
Hi @monsieurleberre, how's it going? Gentle nudge on this, would like your hear your comments. Thank you.

monsieurleberre:
Thanks @ayanuali, and sorry, this sat in my inbox far too long.

The revised model answers both my questions. Separate Script () tests, the adversarial submission on its own with `submitMustFail`, reporting done outside the ledger. That works, and I've got nothing more to add to it 👌 

Three things while you're in there. 
- The earlier sections still say this doesn't exist today, and that now reads oddly against the scope you just described, so it probably wants to change with the rest.
- M2 AC3 runs both testnet parties under Sigli, which tests the harness fine but skips the more interesting case involving two organisations who don't trust each other.
- The 3-month numbers also feel light for the amount, though that's a committee question, not mine.

 I'm one voice in the SIG and my part was only ever the technical read, which is now settled as far as I'm concerned.

ayanuali:
@monsieurleberre thank you, and no problem on the timing. Pushed the revision:
1. Execution model. WorkflowScenario and RunScenario are gone. Scenarios are top-level Script () tests run independently, the adversarial action is its own submitMustFail submission followed by post-rejection state assertions, and reporting is handled by an external runner, not on-ledger. Layer table and a worked scenario are in Technical Architecture.

2. Framing. New "Prior Art" section stating that submitMustFail and Splice's checked variants already exist and are the right tools, and that DAWE builds on them. The "does not exist" claims and the "off-chain assertion code that can be bypassed" line are gone.

3. PolicyConfig. Narrowed to a test-fixture and application-binding interface. Each binding now names the specific policy action and expected rejection.

4. M2 AC3. The settlement suite runs across two parties at two separate organisations, so the authorisation boundary is between parties that cannot sign for one another, with a disclosed fallback if no external counterparty is secured in time.

Since your technical read is settled, what moves this out of Incoming? My understanding is that the champion takes it to Ready for Vote. Happy to raise that with 5North directly, or to answer anything else the SIG needs first.

Thanks!

PR #401, opened 2026-06-01
Proposal: Daml Package Registry

vlady-kotsev:

**Proposal file:**  [proposals/daml-registry.md](proposals/daml-registry.md)

---

## Summary
This proposal outlines the design and implementation of a package registry as an **extension to the existing Daml Package Manager (`dpm`) tool**. The registry enables Daml developers to publish, discover, and consume reusable Daml libraries through a centralized, reliable infrastructure — all from within the `dpm` CLI they already use.

---

## Checklist

---

### DIFF
--- a/proposals/daml-registry.md
+++ b/proposals/daml-registry.md
@@ -1,13 +1,27 @@
-# Daml Package Registry
+## Development Fund Proposal
 
->**Author:** Vladislav Kotsev<br/>
->**Status:** Submitted <br/>
->**Date**: 28.05.2026 
->
+**Author:** Vladislav Kotsev
+**Status:** Submitted
+**Created:** 2026-05-28
+**Label:** daml-tooling
+
+**[Champion](https://github.com/canton-foundation/canton-dev-fund/blob/main/sig-directory.md):** Need Champion
 
 ---
 
-# Rationale and Motivation
+## Abstract
+
+This proposal outlines the design and implementation of a package registry as an **extension to the existing Daml Package Manager (`dpm`) tool**. The registry enables Daml developers to publish, discover, and consume reusable Daml libraries through a centralized, reliable infrastructure — all from within the `dpm` CLI they already use.
+
+The system follows a source-distribution model: authors publish Daml source code, and compilation happens on the consumer's machine using the existing `dpm build` pipeline. No new compiler or build system is introduced; the registry adds registry-aware commands and orchestration on top of `dpm`'s existing capabilities.
+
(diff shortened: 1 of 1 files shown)


Jatinp26:
Some thoughts in my review:

- Overall, I think it's good to be sent for core review. (adding tag for core review)
- Also, could you please share more on how does this relate to the existing dpm OCI registry model? This proposal builds a source-distribution registry (Daml source code, not compiled DARs/OCI images) which is architecturally distinct from the OCI model. Might be helpful to add more info around this.
- Also, have you considered making it into a dedicated dpm component instead? Here's info: https://forum.canton.network/t/dpm-components-extend-the-canton-developer-stack/8822

vlady-kotsev:
As you mentioned the idea is to build source-distribution registry, instead of shipping pre-build artefacts. The goal is to handle compilation on the host machine, to have transparency over the code your import (no black boxing) and to permit compilation with different versions of `damlc` on the fly.

Yes, the Daml registry tool is planned to be integrated in the existing `dpm` tool.

marius080:
Thanks @Jatinp26, @vlady-kotsev. I wanted to add my 2 cents and possibly clarify on the points you are making.

The OCI path in dpm today is about shipping compiled packages as a DAR or a component and people pull the pre-built artifact. That works well when the thing you're distributing is ready to run.
What we're after here is a bit different. We want to make it easy to reuse and share libraries, so we ship source instead of built DARs. A few reasons that matters to us:

- You can actually read the code you're importing instead of trusting a pre-built binary.
- Since it compiles on your machine, the same library can be built against whatever damlc/SDK version you're on. A pre-built DAR is stuck with whatever compiled it.
- The lockfile plus checksums mean everyone (and CI) gets the exact same build.

So we don't see it as competing with the OCI model. We think they cover different needs and can live side by side. Nothing here touches or replaces the OCI path.

On the component idea. We agree that's the right way to do it, and thanks for the link. The components post even lists package registry integrations as a use case, so shipping the CLI side as a proper dpm component is exactly the kind of extension to dpm we had in mind. You would install it like any other component and it plugs into the tooling you already use.

We'll update the proposal to spell out the OCI versus source difference and to say the client ships as a dpm component. Let me know if anything else feels off.

marius080:
@mziolekda 
@hrischuk-da 

Hi to both of you! 
I would like your help in getting this PR reviewed since it's been stuck for a while. Please reach out to me if you have any questions.

Thanks,
M

Jatinp26:
@marius080 the proposal is already under `Ready to vote` so once the current proposals get voted then the current proposals under `ready to vote` will come up automatically.

Jatinp26:
@vlady-kotsev The proposal needs to rescope based on what's already been approved to be built under the proposal by Obsidian for package management at https://github.com/canton-foundation/canton-dev-fund/pull/533

PR #130, opened 2026-03-23
Proposing Daml Package Analyzer

jordan-s-certora:
# Development Fund Proposal: Daml Package Analyzer

**Author**: Certora  
**Status**: Draft  
**Created**: 2026-03-19

---

# Abstract

We propose a lightweight static analysis tool that inspects a Daml project to compute and present information about cross-package interactions in a developer-friendly way. Unexpected references across packages can increase the risk of misuse of sensitive functionality and broaden the attack surface. Being able to precisely compute the dependencies gives users a way to inspect it statically and redesign the system if needed.

---

# Specification

## 1. Objective

  As it stands today, Daml security researchers, developers, and node operators who vet packages must manually analyze packages used in a Daml project to understand how they might interact. This leads to longer security reviews and development time. For example, if package `A` can exercise choices or create templates from package `B` , it might give `A` control over money movement, approvals, etc. which may be undesirable. Examples of behavior that the tool should prevent:  
  1. A `.dalf` cannot be used to steal assets  
  2. A `.dalf` does not negatively affect any app from another provider

  We propose a tool to statically analyze a Daml project and produce a visual output that will show the various ways packages can interact with each other.  The output will also be available as a JSON file so that other tools can leverage its output, similar to the [graphviz](https://graphviz.org/) tool

  The goal is to help auditors and other users analyze a `.dalf` and precisely identify all places in the source code where a package has the authority to call into another package, including the relevant source file and line number for each interaction. This analysis will be for the entire dependency tree or to a pre-specified depth. This will help the auditor focus on the most security critical areas that may inadvertently delegate authority.

## 2. Implementation Mechanics

  The input to the tool will be a project’s compiled `.dar` which contains the main package, its source code, and its dependencies. To a first approximation, the tool will:  
  - Inspect the `.dar` file and identify all the packages used  
  - For every package, identify every location where the code exercises choices defined on a template or interface from another package.  
  - Additionally, report cross-package template usage as a sign of dependency.  
  - For each such interaction, report:  
   1. the source file path  
   2. the exact line number (and column range where available)  
   3. the calling package and package version  
   4. the referenced package/interface and the version  
   5. the template/interface and choice involved  
   6. If it is a non-consuming or consuming choice  
  -  Additionally, report where there are templates that implement specific interfaces  
  -  Aggregate the results and present the information in both a structured machine-readable JSON form and an intuitive graph-like visualization. The structured output will allow auditors to navigate directly from an interaction to the relevant source location.

## 3. Architectural Alignment

[...]

PR #49, opened 2026-03-05
Development Fund Proposal for Maintenance of Daml Open Source

bame-da:
## Development Fund Proposal Submission

**Proposal file:**  https://github.com/digital-asset/canton-dev-fund/blob/dffd5c99d8bfc66d64289c1b997aaea87063acaa/proposals/proposal-daml-open-source.md

---

## Summary
This proposal outlines a grant for Digital Asset to continue its role as the primary maintainer of the Daml and Daml SDK Open Source repositories. It requests a recurring monthly grant of 1,000,000 Canton Coin (CC) to support the ongoing engineering efforts required to ensure the security and stability of the Daml SDK as developed in related open source projects, and allow for its efficient evolution. The specific components include, but are not exclusive to the Daml Smart Contract language, its compiler and IDE, the dpm CLI assistant, client libraries and related code generation, the Participant Query Store (PQS), Daml Shell, Daml Script, and all related documentation. This proposal establishes a single, recurring "maintenance" milestone commencing in March 2026\.

---

## Checklist
- [x] Proposal file added under `/proposals/`
- [x] Milestones and funding amounts defined
- [x] Acceptance criteria included
- [x] Alignment with Canton priorities described

---

## Notes for Reviewers
(Add anything the Tech & Ops Committee should pay attention to.)

PR #327, opened 2026-05-13
Proposal: Add DPM Trace source-aware debugging

djolertrk:
## Development Fund Proposal Submission

**Proposal file:**  
`/proposals/dpm-trace-source-aware-debugging.md`

---

## Summary

We propose to build `dpm trace`, an open-source DPM plugin for inspecting and debugging Canton transactions.
A developer should be able to start from an already committed update id, run a single CLI command, and get a readable transaction trace.
From there, the tool adds interactive inspection and source-aware debugging workflows. To make that reliable, we also propose compiler-generated debug metadata so transaction traces and debugger steps can link back to Daml source.

## Proof of Concept Implementation
- `dpm trace`: https://github.com/walnuthq/dpm-trace
-  Initial `damlc --debug-info` support branch: https://github.com/walnuthq/daml/tree/feature/debug-info 
---

## Checklist
- [x] Proposal file added under `/proposals/`
- [x] Milestones and funding amounts defined
- [x] Acceptance criteria included
- [x] Alignment with Canton priorities described

---

## Notes for Reviewers
We are open for discussion.

PR #494, opened 2026-07-04
Proposal: DPM Debug — Visual Debugging Plugin for Daml on Canton

smehrjerdian:

**Proposal file:**
[/proposals/2026-07-Lazer-dpm-debug.md](https://github.com/LazerTechnologies/canton-dev-fund/blob/proposal-dpm-debug/proposals/2026-07-Lazer-dpm-debug.md)

---

## Summary
`dpm-debug` is a DPM plugin bringing visual transaction debugging, structured test output with assertion diffs, and a searchable debug log panel to the Daml developer experience on Canton — the #1 gap cited in the Foundation's DX survey. Lazer Technologies requests 1,000,000 CC (~$140,000) over 16 weeks across 3 milestones (400K/400K/200K CC), released MIT-licensed as a common-good developer tool.

---

## Checklist

---

## Notes for Reviewers
Runtime debugging complements (does not overlap) Certora's compile-time static analysis (#130) and extends Foundation-funded DPM (#49) via its plugin architecture. Champion: in discussion with the Tech & Ops Committee. Supporting strategy deck: https://lazertechnologies.github.io/canton-debugger-proposal/

### DIFF
--- a/proposals/2026-07-Lazer-dpm-debug.md
+++ b/proposals/2026-07-Lazer-dpm-debug.md
@@ -16,7 +16,7 @@
 
 `dpm-debug` is a DPM plugin that brings visual transaction debugging, structured test output, and a debug log panel to the Daml developer experience on Canton. The Canton Foundation's own developer experience survey flagged visual debugging and transaction visibility as the top missing tools in the ecosystem, and Foundation feedback has repeatedly identified a high-quality debugging experience as the single biggest gap on top of Canton.
 
-This proposal requests **750,000 CC (~$112,500 USD)** over **12 weeks** across **3 milestones** to design, build, and release `dpm-debug` 1.0 as an open-source (MIT) plugin for DPM.
+This proposal requests **1,000,000 CC (~$140,000 USD)** over **16 weeks** across **3 milestones** to design, build, and release `dpm-debug` 1.0 as an open-source (MIT) plugin for DPM.
 
 ---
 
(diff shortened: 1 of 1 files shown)


monsieurleberre:
👋 Hi @smehrjerdian, thanks for the proposal! This looks like a really nice addition to the DPM ecosystem.

I was wondering how this fits alongside PR #327 (Walnut's dpm trace), which was submitted in May and is also looking at building a DPM transaction tree visualizer? 

It seems like `dpm-debug` focuses on local dev-time test runs while `dpm-trace` targets live Ledger APIs, so it would be nice to get a mention of how they compare or if you see any crossover in the Web UI layer. 

Thanks 🙏

smehrjerdian (commented):
Thanks @monsieurleberre for the thoughtful questions, it prompted us to dig into #327 properly, and we've now revised our proposal to make the relationship explicit rather than leaving it implied.

One small correction to the framing first: we don't think local vs. live Ledger API is quite the right axis. A local Canton participant exposes the same Ledger APIs as a deployed node, and dpm trace will be just as useful locally. The real distinction is layer, not environment.

Having reviewed #327 (and the debug-metadata draft in #382) closely, we think the Walnut team is building exactly the right foundation — the CLI trace, portable trace bundles, interactive debugger, and especially the compiler-emitted debug metadata standard are the hard, low-level primitives the ecosystem needs, and their compiler-tooling background makes them the right team for it. We have no interest in duplicating any of that, and we've removed the pieces of our original scope that edged toward it (snapshot/replay is gone — Walnut's trace bundles are the right answer there).

What we've revised our proposal to be: the visual workbench layer on top of that foundation — a DPM plugin that launches a local web app (think Remix in the EVM ecosystem): a participant explorer showing deployed packages, templates, and visible contracts; a live transaction feed; interactive transaction trees; and visual test reports with assertion diffs. On the Web UI crossover specifically: #327 lists a web UI only as an unfunded follow-on, and #382 explicitly names "visualization and test-reporting tools" as intended consumers of the metadata standard. That's us. We commit to rendering dpm trace's normalized JSON trace artifacts and trace bundles natively, and to consuming the #382 metadata for source display rather than inventing a parallel format — with graceful fallback to direct Ledger API queries so nothing is hard-blocked on sequencing.

Our intent is to stay firmly in our lane: Walnut builds the primitives, we make them visible and approachable for the median application developer. We'd welcome an introduction to the Walnut team to align on the artifact formats — a shared standard here benefits everyone regardless of which proposals move forward.

The revised proposal is in the latest commit on this PR, and the [strategy deck](https://lazertechnologies.github.io/canton-debugger-proposal/) has been updated to match.

Jatinp26:
@smehrjerdian 
- I am NOT championing this proposal, kindly remove that.
- How's this gonna fit with current usecase AFTER wallnut's tooling?
- Where's the current ecosystem demand for this?

Need this info to proceed.

PR #7, opened 2026-02-20
feat(lsp): add proposal for DAML LSP

0xfourzerofour:
## Development Fund Proposal Submission

**Proposal file:**  
[proposals/daml_lsp.md](https://github.com/canton-foundation/canton-dev-fund/pull/7/changes#diff-86e6cb2b3b5d7044c9b542f5dd4177ed2dc42c3835c4fcb36d7ebe7676afee60)

---

## Summary
This proposal seeks funding to build a native Language Server Protocol (LSP) implementation for DAML, the smart contract language used on the Canton Network. Today, DAML developers rely on the Haskell LSP as a partial workaround, which lacks DAML-specific features such as template and choice awareness, contract key validation, and rights-checking diagnostics. A purpose-built DAML LSP will provide first-class editor support including intelligent autocompletion, real-time diagnostics, go-to-definition across DAML modules, and inline documentation, significantly lowering the barrier to entry for Canton smart contract development and improving productivity for existing developers.

---

## Checklist
- [x] Proposal file added under `/proposals/`
- [x] Milestones and funding amounts defined
- [x] Acceptance criteria included
- [x] Alignment with Canton priorities described

---

PR #531, opened 2026-07-14
Proposal: Daml Language Server Modernization

heitor-lassarote:

**Proposal file:** [proposals/lsp_proposal.md](https://github.com/serokell/canton-dev-fund/blob/heitor-lassarote/serokell-lsp-proposal/proposals/lsp_proposal.md)

---

## Summary

In [PR #7](https://github.com/canton-foundation/canton-dev-fund/pull/7), @0xfourzerofour proposed rewriting the Daml language server from scratch. The author went inactive and the proposal was rejected. We'd like to revive that effort, with a different approach.

Rather than a rewrite, we propose forking Haskell Language Server (HLS), the de facto production-grade language server for Haskell, and adapting it to Daml on top of Damlc (itself a fork of GHC), adding the Daml-specific changes needed to bridge the gap with stock HLS. Daml Studio's language server today is Digital Asset's own 2019 fork of Ghcide, predating HLS, and it has not kept pace with the LSP feature set HLS has since built (rename, formatting, call hierarchy, and more). It also has no officially supported path beyond VS Code, despite some unofficial community integrations for other editors. Building on HLS' plugin architecture addresses both gaps directly, while reusing the same Ghcide extension points that Daml's own fork has relied on successfully since 2019.

Our team includes certified Daml developers, GHC maintainers, designers of the first version of the Cardano node and wallet, general Haskell tooling engineers, and contributors to the LIGO, Michelson, and Motoko tooling ecosystems.

---

## Checklist
- [X] Proposal file added under `/proposals/`
- [X] Milestones and funding amounts defined
- [X] Acceptance criteria included
- [X] Alignment with Canton priorities described

---

## Notes for Reviewers

Notes for Reviewers

- Milestone 1 is scoped as a short investigation phase to close open technical questions before committing to the heavier Milestone 2/3 build.
- HLS 1.8.0.0 is proposed as the specific fork point: the last HLS release to support GHC 8.8.4 (close to Damlc's current compiler version: GHC 8.8.1), and already a mature codebase by the time it shipped in September 2022. Subject to review in Milestone 1.
- Acceptance criteria are adoption-based (developer migration count, multi-editor uptake, community PRs/issues, adoption enablement) rather than artifact-delivery-based, per CIP-0100's guidance to weight proposals toward ecosystem utility.
- Vim/Neovim support is scoped as its own milestone within this proposal, rather than a separate grant, to keep it accountable to this proposal's own acceptance criteria.

---

In https://github.com/canton-foundation/canton-dev-fund/pull/7#issuecomment-4303306132, some questions were asked regarding the old LSP proposal. Here we'd like to answer them:

1. AST Parser: As we plan to fork HLS, we would always be up-to-date with whatever revisions of Damlc are pinned to it, provided the fork links against DA’s own patched GHC/`ghc-lib`.
2. Type-checking: The same. HLS uses the GHC API for syntax and types, so we’d use the compiler as the source of truth.
3. Daml Versions: The plan is to decouple `damlc ide` from `damlc` and use HLS’ strategy to build the new language server using specific Damlc versions and use its API. Since the language server will be tied to Damlc internally, making changes to it should be low-risk, but it also doesn’t come for free if the API changes. Historically, Damlc's vendored `ghc-lib` frontend has been pinned to GHC 8.8.1 unchanged since the project's inception in 2019, which justifies this reasoning.
4. Acceptance Criteria: As we would use the same parser Damlc already uses, we can rest assured that all syntactical constructs can be parsed with no extra effort. However, the syntax highlighting files (TextMate grammars, and the proposed (Neo)Vim grammars) would still require manual intervention. Unfortunately, that is difficult to test and will require synchronization from the core .y grammar maintainers and the language server maintainers.
    1. A lower-effort solution to keep the editor grammars in sync is to use a tool such as https://gitlab.com/ligolang/ligo/-/tree/dev/lib/ligo_syntax_highlighting (a tool we helped to maintain which we may fork and rewrite in Haskell), minus Emacs grammars, but it wouldn’t help with keeping them in sync with the core grammar.
    2. Alternatively, we may investigate about reading Happy source files, or possibly its generated debug output, and write a tool for it, but it’s a high-effort project.

### DIFF
--- a/proposals/lsp_proposal.md
+++ b/proposals/lsp_proposal.md
@@ -1,0 +1,341 @@
+# Daml LSP Development Fund Proposal
+
+**Author:** Heitor Toledo Lassarote de Paula, on behalf of [Serokell OÜ](https://serokell.io)
+
+**Status:** Submitted
+
+**Created:** 2026-07-14
+
+**Label:** daml-tooling
+
+**[Champion](https://github.com/canton-foundation/canton-dev-fund/blob/main/sig-directory.md):** Need Champion
+
+---
+
+## Abstract
+
+This proposal aims to replace the legacy Ghcide-based Daml language server with a modern fork of Haskell Language Server (HLS), tailored specifically for Daml. This upgrade will bring Daml developers new capabilities that streamline development, help onboard new developers, move Daml Studio onto a newer and battle-tested foundation, and provide a platform for future Daml- and Canton-specific features.
+
+---
+
+## Specification
+
+### 1. Objective
+
+Daml Studio’s language server is still being built on Digital Asset’s own 2019 fork of Ghcide, a project that Digital Asset itself originated and handed over to the Haskell community. The fork of Ghcide is essentially frozen at around the time [the project was handed over](https://blog.digitalasset.com/blog/handing-over-ghcide-to-the-haskell-community) and [joined with the Haskell IDE Engine](https://neilmitchell.blogspot.com/2020/01/one-haskell-ide-to-rule-them-all.html) project in January 2020 to found the Haskell Language Server.
+
+The Canton Development Fund explicitly states to [extend what already exists](https://github.com/canton-foundation/canton-dev-fund/blob/38217826a7c0ca1c2b456b1bd5623649ecff6f71/proposals/_template.md#rationale) by default. In the upstream Ghcide, there is still [a vestigial comment](https://github.com/haskell/haskell-language-server/blob/048cb169ce269e46015afcb9631e9378400bb272/ghcide/src/Development/IDE/LSP/LanguageServer.hs#L6-L7) requesting to keep it in sync with `DA.Daml.LanguageServer`; a resulting scar tissue showing that a drift happened and Daml never came back to the lineage it helped to start. [Another comment](https://github.com/digital-asset/daml/blob/139b01b2c6a07c61d32507eaee0a16b5d76a6a09/sdk/compiler/damlc/lib/DA/Cli/Damlc/Command/MultiIde/Forwarding.hs#L144) also makes it clear that the Ghcide fork has fallen behind and that many capabilities are now supported by HLS.
+
+In face of this, we propose to fork HLS, a production-grade battle-tested language server for Haskell to create a new Daml language server with similar features. The objective is to provide Daml developers with a modern IDE experience built from existing HLS features while keeping existing Daml Studio features. Moreover, we wish to untie the Daml experience from Visual Studio Code exclusively by creating thin Daml IDE wrappers for Vim (and Neovim).
+
+### 2. Implementation Mechanics
+
+The [Language Server Protocol](https://microsoft.github.io/language-server-protocol/specifications/specification-current/) (LSP) is a JSON‑RPC protocol that lets editors communicate with a language server to provide real‑time diagnostics and code intelligence (navigation, refactoring, formatting, etc.). Today, `damlc ide` is a Haskell process that implements an LSP backend for Daml, based on a 2019 fork of Ghcide and extended with Damlc-specific rules. Those rules are compiled against Daml’s fork of GHC 8.8.1. Haskell Language Server (HLS) is built on the same Ghcide-derived architecture; because Daml is Haskell-shaped and Damlc is itself a GHC fork, we expect Daml to be a good fit for an HLS-based implementation. In other words, this work retargets HLS to Daml rather than rewriting the language server from scratch.
+
+We currently identify the following key components:
+
+1. **Damlc**: Daml’s compiler, implemented in Haskell and based on a fork of GHC 8.8.1. Where needed, we will port existing mechanisms to support reliable IDE integration rather than reimplementing them.
+2. **Damlc IDE (damlc ide)**: a retargeted and streamlined fork of Haskell Language Server (HLS), including a curated set of plugins. It will be extended with Daml-specific handlers for custom LSP methods.
+3. **Ghcide IDE state and Shake incremental build graph**: we expect these to remain largely unchanged, subject to confirmation in Milestone 1.
+4. **Compiler-hook extension points**: `optPreprocessor` (AST-level desugaring, syntax, and lint) and `optGhcSession` (session loading driven by `daml.yaml`). Both appear preservable, but Daml’s current preprocessor takes extra arguments (including `GHC.DynFlags`) not present in our likely HLS baseline’s version of this field, so some adapter work may be needed here. We will validate the exact shape of the scope in Milestone 1.
+5. **Daml Studio**: the editor client layer, delivered primarily as a Visual Studio Code extension. The TypeScript code covers functionality that is not easily expressed through standard LSP, such as the web views for table and transaction results. We also plan to provide thin LSP-based integrations for Vim/Neovim, alongside syntax highlighting.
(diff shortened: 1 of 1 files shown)


heitor-lassarote:
Hi, @hrischuk-da, @nycnewman, @mziolekda, @paulbrauner-da, I'm Heitor from Serokell. We just submitted this development fund proposal to modernize the Daml language server but our proposal got automatically closed pending a champion. The proposal is ready, just otherwise needs the champion field.

Short version, Daml's 2019 Ghcide fork hasn't kept up with upstream changes, straying from the LSP feature set that Haskell Language Server has built since then. This is a proposal to fork HLS (possibly version 1.8.0.0) and adapt it on top of Damlc, closing that gap, and also to write a Vim/Neovim plugin.

Would you be willing to champion this proposal and reopen it? I would be more than happy to walk through any part of it if you find it useful.

alterroman:
I would also like to ask @shaul-da here. Shaul, any chance you could recommend the best way for us to proceed and find a champion for our grant proposal? Thank you in advance!

heitor-lassarote:
Hi @hythloda, tagging you since you reviewed and closed #7, and this proposal responds directly to the concerns raised there (see the Notes for Reviewers on the PR description). Given your role, would you be able to review this proposal, or point us to a Tech & Ops Committee member who could champion it?

@Denend, tagging you too since you raised the technical questions on #7. Both are addressed directly in this proposal and in the Notes for Reviews on the PR description. Happy to discuss further if you have thoughts.

proposals/lsp_proposal.md
```diff
@@ -0,0 +1,341 @@
+# Daml LSP Development Fund Proposal
+
+**Author:** Heitor Toledo Lassarote de Paula, on behalf of [Serokell OÜ](https://serokell.io)
```
hrischuk-da, line 3: @heitor-lassarote thank you for taking the time to write the proposal.  We recognize there is work to be done here.  The first and necessary step is to migrate to a modern Haskell Language Server.  This requires moving to a recent GHC version.   The rest is downstream of that.   Is it worth having a discussion?

heitor-lassarote: Hi, Curtis, thanks for reading the proposal and raising this. Yes, this is definitively worth having a discussion.

We agree that the GHC and HLS versions are linked and can't be separated, since HLS links directly against the real GHC library. Damlc builds on a forked GHC 8.8.1, and many changes have been done since then, specially in GHC 9.2 which seems to have some big changes (like changes to the module layout and exactprint scheme). This is why we proposed HLS 1.8.0.0 (targeting GHC 8.8.4), which also supports pre-9.2 GHC versions. While researching this proposal, we independently found that targeting a more recent HLS version does mean moving Damlc's GHC fork forward first. But before we turn this into a new milestone, there are some things to discuss.

What do you mean by "recent GHC version"? Are you thinking about a specific release, or the latest (9.14.1 as of now)? We believe that staying pre-9.2 should be more contained work, but moving forward should be a much bigger effort.

And do you know whether the Canton Foundation, Digital Asset, or other organization already have a plan to upgrade Damlc off GHC 8.8.1, or is it something to be done purely within this proposal's scope?

With these answers we can better size the scope of a new milestone gating the others. Even as [HLS docs say](https://github.com/haskell/haskell-language-server/blob/76af2b9341fd76f04f741334c83d2623bbce2478/docs/support/ghc-version-support.md#L146-L151):

> `haskell-language-server` (HLS) is highly tied to the GHC API. This imposes a high maintenance cost:
>
> * The codebase is littered with conditional logic
> * We own auxiliary packages to support older versions of GHC
> * CI has to cover all the supported versions

Also, would a call work better, or do you prefer to continue discussing this just here in GitHub?

Jatinp26:
@heitor-lassarote 
- what's the fallback if the Windows build can't be fixed without a GHC version bump? Is Windows supporting a hard acceptance criterion for M2?
- The TTG annotation work for template/choice/interface blocks is correctly flagged as scope-at-risk. If it doesn't fit M3's budget, the fallback is shipping without it.
- The Milestone 6 maintenance budget is deliberately TBD.
- Given that this proposal is essentially an infrastructure project add on, have u chatted with anyone in engineering / DA yet on this?
- This needs work with the core engg team and buy-in.

heitor-lassarote:
> @heitor-lassarote
> 
> * what's the fallback if the Windows build can't be fixed without a GHC version bump? Is Windows supporting a hard acceptance criterion for M2?
> * The TTG annotation work for template/choice/interface blocks is correctly flagged as scope-at-risk. If it doesn't fit M3's budget, the fallback is shipping without it.
> * The Milestone 6 maintenance budget is deliberately TBD.
> * Given that this proposal is essentially an infrastructure project add on, have u chatted with anyone in engineering / DA yet on this?
> * This needs work with the core engg team and buy-in.

Hi @Jatinp26, thanks for reading the proposal. Point by point:

- About Windows: the short answer is yes: Windows CI stays a hard M2 criterion. We don't yet know for certain whether the segfault would reproduce on our fork. We'd rather be upfront about that than assume it away.
    - We looked back at the actual upstream bug in detail. It's a bug in GHC's Windows code generator, not something specific to `floskell` or to which GHC compiled HLS. It fires on large or complex Haskell modules when built with optimization on Windows. The original report was closed unfixed, but only because GHC 8.8 was already end-of-life by then. A similar crash showed up again later, on GHC 9.5, on an unrelated package, and it was fixed in GHC 9.6.1. But the fix doesn't apply clearly to us, since Damlc is on 8.8.1 and the bug only applies to 9.5 or later.
    - We don't plan to integrate a formatter plugin, so we won't touch the module that triggered the original crash. But the bug is about large or complex modules in general, so it could still pop up somewhere else.
    - This connects to the GHC-version conversation we're also having with @hrischuk-da above. There isn't a ready-made fix to reuse for our situation, so our real fallback path is:
        1. Confirm whether it happens at all. If nothing happens, we remove this risk before M2. This is what the Windows build spike is about below.
        2. If it does crash, we try a compiler swap. `ghc-lib` decouples which GHC compiles our tool from which GHC our tool implements, so building with a newer GHC is something we can attempt in a few days. There's no evidence this fixes anything, but it's a cheap thing to try before committing to heavier work.
        3. If neither of the above works, we patch it ourselves. We control the full stack, and the team includes GHC maintainers, but this is open-ended: real work with no fixed size until we know what we're looking at.
    - About how would we know if we'd fixed it: One of the bug reports shows that retrying an incremental build looks nondeterministic (a module fails, you rerun, it passes, a different module fails instead), but a full, clean, fully-optimized rebuild reproduces the crash reliably. So our Windows testing will always be a clean build.
    - To address this risk early rather than at M2 acceptance, we propose amending M1 to include an explicit Windows build spike. We would boot the fork skeleton on Windows in M1, with fully-optimized clean builds, before the M2/M3 budget is committed. Happy to push that amendment to the proposal text if the Committee finds it useful.
- About the TTG work: yes, that's the intended reading. If M1 sizing shows it doesn't fit M3's budget, we ship without it, with explicit disclosure and the same amendment path.
- About the TBD in Milestone 6's budget: also correct, and deliberate: a responsible maintenance number depends on post-release facts (issue volume, SDK release cadence, whether the server is adopted into the official SDK distribution). We're committing to the maintenance now; we propose fixing structure and budget with the Committee around M3–M4, per CIP-0100's quarter-on-quarter maintenance milestone option.
- About DA engineering engagement: We want to correct something first: in this issue, in the comment above yours, @hrischuk-da, from the Daml Language & Developer Tooling SIG, left a comment about the same GHC/HLS version question above, which I've replied to. But that thread's been quiet for about a month since and nothing substantial happened, so we fully agree more engagement is essential.

Given this, there are two things we would like to request to you:

1. We're happy to amend M1's acceptance criteria so the decision document requires review and sign-off from DA core engineering, alongside the champion/Committee. The technical direction then gets approved by the owners of the code before the heavier M2/M3 budget is committed.
2. Could you help us connect to the correct owner on the tooling/damlc side? We could have a 45-minute call with the technical plan to establish whether the direction lands. If it goes well, we'll welcome your guidance on the champion path, which is the current blocker for this PR.

We can make ourselves available any day this week or next.

proposals/lsp_proposal.md
hrischuk-da: @heitor-lassarote We do not think building on top of GHC 8.8.1 is the right approach and cannot support this grant proposal as is.  We would like to discuss the alternative with you.   Please reach out to me at curtis.hrischuk@digitalasset.com

heitor-lassarote: Thank you for getting back to us on this. I've sent you an email.

PR #574, opened 2026-07-16
Proposal: Canton Mobile SDK (Swift + Kotlin)

sunstrike228:

**Proposal file:**
./proposals/2026-07-VertexPoint-canton-mobile-sdk.md

---

## Summary

Canton Mobile SDK: native Swift (iOS) and Kotlin (Android) libraries for building on Canton from the two platforms end users hold. Covers the JSON Ledger API v2 client surface, on-device key management for external parties (secure element, biometric-gated interactive-submission signing), token standard V1 and V2, pre-built Splice bindings, and a native implementation of the CIP-0103 wallet connectivity surface over WalletConnect.

Total ask 1,150,000 CC, weighted 66% toward adoption: 600,000 CC pays only against independent Featured Apps running mobile clients on the SDK in production on mainnet. Engineering (390,000 CC) is gated behind acceptance milestones with a working proof of concept at Milestone 1. Apache-2.0 from the first release.

---

## Checklist

---

## Notes for Reviewers

- Champion: not assigned yet, submitting as Need Champion. Happy to walk any committee member through the proposal.
- The funding shape follows Peaceful Studio #46: per-event adoption tranches, a security-review pass-through held outside the engineering envelope (capped at 130,000 CC), and an extended adoption window requested as an explicit exception (12 months; #46 carries 18).
- Prospective adopter teams from the wallet side of the ecosystem are being invited to comment directly on this PR so the review is informed by teams that intend to build with the SDK.
- The signing-scheme question (secure elements hold P-256, current documentation examples use Ed25519) is addressed head-on: Milestone 2 demonstrates onboarding and settlement with both schemes before the token-standard milestone opens.


vbloher:
Great initiative. I'm currently building a DeFi project and mobile apps are on our roadmap, so having a proper SDK with JSON Ledger API and CIP-0103 support would save us a significant amount of integration work. Happy to help review milestone deliverables along the way.

pmensik:
This addresses a real gap in the Canton ecosystem. A reusable native mobile SDK covering Ledger API access, external-party signing, token standards, and wallet connectivity could save mobile teams substantial duplicated integration work.

I also like that most of the funding is gated behind demonstrated production adoption rather than code delivery alone. The scope across two platforms is ambitious, especially around secure key custody and signing so the early milestones should be evaluated strictly. Overall, this looks like a valuable and thoughtfully structured proposal, Ruby Nodes definitely support it!

anand-qasara:
We (Qasara) are building Cove, a non-custodial REST gateway on Canton: prepare, sign, broadcast, and we never hold keys. Your mobile SDK and Cove sit at complementary layers. Your SDK holds the key in the secure element and signs on-device; Cove prepares the transaction and broadcasts it on the backend. In CIP-0103 terms your SDK is the on-device (synchronous) end and Cove is the server-side (asynchronous) end of the same standard, so a dApp written once could reach either.

We would use this. The seam we care most about is the prepare to sign to execute handoff: payload shape, hashing scheme, and what the phone shows the user before signing. Happy to be a design partner there, and to compare notes on the signing side, including the open question your M2 flags of whether the Global Synchronizer accepts P-256 external-party keys or needs Ed25519.

Jatinp26:
@sunstrike228: 

- Having a Kotlin and Swift can be beneficial for some unintroduced areas of development for sure.
- I liked how most of it is based on adoption, that is indeed great as a proposal.

Technical questions:
- The proposal cites `/v2/parties/external/generate-topology` and `/v2/parties/external/allocate` as the onboarding endpoints and AFIK These are Splice Validator API endpoints not canonical Canton JSON Ledger API. For private Canton deployments without a Splice validator, external party onboarding goes through the Admin API's `TopologyManagerWriteService` Which path does the SDK support and how does it handle non-Splice deployments?
- External party onboarding requires a namespace signing key distinct from the protocol signing key. The proposal describes generating protocol signing keys in the Secure Enclave. How is the namespace key generated, stored, and used? Does the SDK use a single key for both namespace and protocol purposes, and if so, is that stated explicitly in the security model?
- Verifying the prepared transaction hash requires implementing Canton's transaction hashing scheme on device. Is this implemented in the KMP core? What's the implementation approach parsing the protobuf and computing the hash according to the spec or something else?

sunstrike228:
@Jatinp26 thanks, appreciate the questions -

On the endpoints, those two are Canton JSON Ledger API v2 rather than Splice. Canton documents the flow in `external_signing_onboarding_lapi.rst`, ships a TypeScript example calling both paths in `examples/14-multisync/src/parties.ts`, and tests it in `ExternalPartyLedgerApiOnboardingTest`. searching hyperledger-labs/splice for `generate-topology` returns nothing. so it is the plain participant path and works with no Splice validator in the deployment. building the transactions manually over the Admin API stays open to operators who want it, but the device does not need that route

On the namespace key, the endpoint takes one public key and derives the party namespace from its fingerprint, so in that flow a single device key is both namespace root and protocol signing key. that is what v1 implements, and you are right that it should be stated in the security model rather than implied, along with the consequence that losing the device loses the party. we will spell it out and add onboarding with a separate recovery key that co-controls the namespace

On hashing, yes, in the KMP core so both platforms use one implementation. it parses the returned `preparedTransaction` and recomputes the hash for the scheme named by `hashingSchemeVersion`, then compares against what the participant sent. CI validates it against vectors from Canton's own implementation, and a scheme bump is treated as a compatibility event. your question applies to onboarding too -the tutorial notes the convenience endpoint is fine when the node is trusted, so we will decode and check the returned topology transactions instead of signing the multi-hash blind

sunstrike228:
@Jatinp26 all three are in the proposal now, pushed in d05f8b6

Jatinp26:
@sunstrike228 the revised proposal is substantially stronger. Many of the technical concerns from the initial review have been addressed properly, but still something need to be flagged here.

- External party onboarding of "with or without a Splice validator" needs specific M2 verification.
   - The proposal says `/v2/parties/external/generate-topology` and `/v2/parties/external/allocate` work "with or without a Splice validator in the deployment."AFIK...external party onboarding goes through `TopologyManagerWriteService` on the Admin API directly nd `/v2/parties/external/` endpoints are Splice Validator API convenience wrappers. Whether they exist on a non-Splice Canton deployment needs verification.
   - Maybe M2's acceptance criteria should include demonstrating external party onboarding against a non-Splice Canton participant (direct Ledger API, no Splice validator in the path). If this succeeds, the claim holds and the SDK genuinely works across deployment types. If it fails, the scope should be narrowed to Splice-based deployments explicitly which is still a valid and useful scope just a different one than what's currently stated.

sunstrike228:
@Jatinp26 thanks, Jatin 

Both endpoints are in Canton's own JSON Ledger API OpenAPI spec and Canton's example `08-interactive-submission/external_party_onboarding.sh` calls them against a plain `./bin/canton` run with no Splice in the path

Your remedy is the better way to close it though, so M2 acceptance will include onboarding against a non-Splice participant over the direct Ledger API. if it fails there, scope narrows to Splice deployments as you describe

Denend:
Which CIP-0103 methods are in scope for v1? The specification lists ten synchronous methods and the registered WalletConnect set is seven, but the proposal enumerates five. Is full conformance a v1 goal or a continuation item?

sunstrike228:
The five in the abstract were just the signing-related ones, not the full scope

The difference between ten and seven is the transport. connect, disconnect and isConnected are things a WalletConnect session already handles, so they are not exposed as canton_ methods. everything else is

v1 covers the whole surface, all ten plus the three events. we test against the OpenRPC spec in splice-wallet-kernel because that is the machine-readable one, and the methods go into the M3 deliverables

PR #642, opened 2026-08-18
TEE covalidator proposal

alexeykoren:

**Proposal file:**  
/proposals/2026-08-FPI-TEE-covalidator.md

---

## Summary
The proposal introduces an open-source Canton co-validator that runs within a Trusted Execution Environment, ensuring the operator cannot read, alter, or sign for the node's state. This delivers significant value to the Canton ecosystem by allowing participants to increase their security through additional confirmations without needing to grant new organizations visibility into their private transaction data.

---

## Checklist

---


proposals/2026-08-FPI-TEE-covalidator.md
```diff
@@ -0,0 +1,138 @@
...
+
+Out of scope: compliance and disclosure logic, analytics integrations, and key-custody services. These are downstream consumers of the open interfaces defined here, not part of this proposal.
+
+**2\. Implementation Mechanics**
+
+**Packaging.** We run a Splice participant and its PostgreSQL store as a single multi-process image inside a TEE Enclave. It is wrapped by a harness running inside an enclave, developed by FPI. It manages enclave access and acts as a proxy for certain operations. The deliverable is that image, host-side setup scripts (docker-compose and Terraform), and documentation; standard host components such as the vsock proxy are documented rather than shipped. A working prototype exists: the enclave-packaged node has run against a simulated Canton network, and we separately operate standard, non-enclave validator and TestNet participant nodes. That operating experience feeds directly into the tooling and runbooks this proposal delivers. Milestone 1 merges the two tracks: the enclave-packaged node synchronizing DevNet.
```
da-tanabe, line 49: We've seen some challenges in running Postgres in the cluster (scale, performance, reliability), and have generally had a much better experience with cloud databases. Is Postgres in the cluster the only option?

da-tanabe: Also curious on how things like backups/restore from historical data should be handled.

proposals/2026-08-FPI-TEE-covalidator.md
```diff
@@ -0,0 +1,138 @@
...
+
+**5\. Software upgrades**
+
+Operators are able to upgrade Splice to newest versions without a complicated process involving direct involvement of a Party. We assume trust to Splice binaries signed by Digital Asset. Based on this assumption, the harness inside an enclave will accept an upgrade command from an Operator and switch running Splice binaries internally. This is done without a Party directly authorising an update. 
+
+**6\. Limited access to the Canton node**
+
+Operators don’t have direct full access to Ledger API to protect Party’s privacy. Access to debug logs is also limited. For cases when selective access is required, e.g. for mapping a Party to a participant \- there is a command in harness that will deliver required call data and signatures to the node. This may include, for example, enabling temporary access to debug logging for operators.
```
da-tanabe, line 144: Would this be a wrapper console, with the original console being hidden? Or do you envision more of an "allowlist" of commands? As new commands are added to canton console, how would the list of "approved commands" be maintained?

proposals/2026-08-FPI-TEE-covalidator.md
```diff
@@ -0,0 +1,138 @@
...
+
+**3\. Architectural Alignment**
+
+Functionally, a TEE co-validator behaves exactly like any confirming participant in the existing co-validation model. It hosts parties, validates its view, and counts toward confirmation thresholds. Protocol, standards, and topologies are unchanged. The difference is the trust boundary: the node's operator sits outside it, so co-validation no longer requires showing another organization the party's transactions. Canton gains cooperative validation that preserves financial privacy. The work fits the fund's stated scope under CIP-0082 (security, reference implementations, critical infrastructure) and operates within CIP-0100 governance.
+
+**4\. Backward Compatibility**
+
+Purely additive. A TEE co-validator joins a party's hosting set like any participant. No protocol changes; no changes required from other nodes; removal is an ordinary topology change. A natural follow-on, outside this proposal's scope, is protocol-level recognition of TEE attestations, so that the network itself, rather than the key provisioner, blesses valid enclave builds. We would be glad to draft that as a separate proposal once this model has proven itself.
```
da-tanabe, line 136: Many types of topology transactions generally need admin access in order to submit, and this admin API access is typically not opened up outside of the hosting company's local network/cluster. How do you envision party topology transactions to be submitted to the node in a secure way?

proposals/2026-08-FPI-TEE-covalidator.md
```diff
@@ -0,0 +1,204 @@
...
+* *Audit key (optional).* The enclave can additionally encrypt state exports to a configured public key. Local key generation ships in the repository, and self-custody is the default; an HSM or an external custodian is a configuration choice.
+
+```mermaid
+flowchart LR
+    subgraph EC2["EC2 PARENT INSTANCE"]
+        subgraph ENC["NITRO ENCLAVE"]
+            direction TB
+            NOTE["no disk · no NIC · vsock only · RAM-resident"]
+            SP["<b>Splice participant</b><br/>validates · confirms"]
+            PG["<b>PostgreSQL</b><br/>in RAM"]
```
thibault-da, line 68: Canton is already quite memory greedy, adding the whole persistence layer in the RAM as well could be massive. Have you thought about what kind of numbers are realistic here?

proposals/2026-08-FPI-TEE-covalidator.md
```diff
@@ -0,0 +1,204 @@
...
+            NOTE ~~~ SP
+            SP ~~~ IK
+            PG ~~~ PK
+        end
+    end
+    NET["<b>CANTON NETWORK</b><br/>counts the confirmations"]
+    ST["<b>EXTERNAL STORAGE</b><br/>ciphertext only<br/>snapshots · exports"]
+    AH["<b>SK_audit HOLDER</b><br/>self-custody (default) · HSM · service provider"]
+    IK -- confirmations --> NET
+    PK -- snapshots --> ST
```
thibault-da, line 81: Are those literal Postgres DB snapshots? Are they automatically and periodically exported?

proposals/2026-08-FPI-TEE-covalidator.md
```diff
@@ -0,0 +1,204 @@
...
+    ST -. restore on restart .-> PG
+    AK -- exports --> ST
```
thibault-da, line 83: What's the difference between `snapshots` and `exports` ?

proposals/2026-08-FPI-TEE-covalidator.md
```diff
@@ -0,0 +1,204 @@
...
+
+**Attestation.** We build reproducibly, publish PCR0/1/2 per release, and provide a verification guide so any third party can rebuild the image from source and check the measurements against a running enclave’s attestation document. Trust in the node reduces to trust in published, reviewable code.  We use Splice binaries signed by Digital Asset as a core building block for reproducible builds. 
+
+**Keys.** Three keys, with three distinct jobs:
+
+* *Canton identity key.* The co-validator's signing key, the identity whose confirmations the network counts. It is provisioned into the enclave and never leaves it. This is what makes the co-validation itself trustworthy: if this key were available outside the TEE, its holder could sign confirmations from anywhere.
```
thibault-da, line 55: A participant node has generally a few different signing keys:
- Root namespace key: derives the identity of the node and controls the namespace (topology state) of the node, including rotating of the 2 keys below
- Protocol signing key: Signs protocol messages (including confirmation responses)
- Sequencer authentication key: used by the validator to authenticate itself towards the sequencer

Can you clarify which of these are you referring to specifically here? I suppose the first one?
Also have you thought about how key rotation of the non-root keys would look like?

rfps/security-assurance-incident-readiness/2026-08-FPI-TEE-covalidator.md
```diff
@@ -0,0 +1,228 @@
...
+    class NOTE note
+    class SP,PG,NET,ST solid
+    class IK,PK,AK,AH key
+    class EC2 outer
+    class ENC encl
+```
+
+**Figure 2** — Three keys, three jobs. The identity key never leaves the enclave; the persistence key returns only to a build with matching measurements; the audit key is an optional, operator-configured interface whose private half sits wherever the operator chooses. The operator of the EC2 parent instance sees ciphertext and vsock traffic only.
+
+**Provisioning.** The party that benefits from the co-validation, typically the asset issuer, provisions the keys. A fresh enclave generates a keypair and produces an attestation binding its public key to the measurements of the running build. The issuer verifies the attestation against the published values and injects the identity and persistence keys, encrypted to the attested public key. An independent infrastructure operator can therefore run the co-validator without ever being able to read its state or sign on its behalf. The same procedure applies to enclaves the issuer runs itself.
```
simran-da, line 103: In our case, the key provisioning (as well as key rotations and software upgrades) would be done by the node operator. The attestations need to be made available to the parties hosted on the node. Would the distribution of the attestations be done by exposing an API from within the enclave, through perhaps a challenge-response protocol, to ensure the attestation is fresh?

rfps/security-assurance-incident-readiness/2026-08-FPI-TEE-covalidator.md
```diff
@@ -0,0 +1,228 @@
...
+
+**Keys.** Three keys, with three distinct jobs:
+
+* *Canton identity key.* The co-validator's signing key, the identity whose confirmations the network counts. It is provisioned into the enclave and never leaves it. This is what makes the co-validation itself trustworthy: if this key were available outside the TEE, its holder could sign confirmations from anywhere.
+
+* *Sealed persistence key.* Enclaves have no persistent storage, so state lives outside as encrypted snapshots. This key encrypts them, and it is released only to an enclave whose measurements match the published build. A new enclave instance restores from the latest snapshot instead of starting from zero, and no human ever holds the key.
+
+* *Audit key (optional).* The enclave can additionally encrypt state exports to a configured public key. Local key generation ships in the repository, and self-custody is the default; an HSM or an external custodian is a configuration choice.
```
simran-da, line 63: If the node operator holds the private (audit) key, what would ensure that it is not misused? is there a way for the affected hosted parties to be notified when a decryption of their data with the audit key is done?

rfps/security-assurance-incident-readiness/2026-08-FPI-TEE-covalidator.md
```diff
@@ -0,0 +1,228 @@
...
+
+**RFP alignment.** This proposal responds to **RFP 23, Validator and Shared Infrastructure Security and Resilience,** in the [2026–2028 Strategic Roadmap](https://github.com/canton-foundation/canton-dev-fund/blob/main/2026-2028-strategic-roadmap.md), which seeks "reusable tools, controls, and standards that improve the security, reliability, availability, and recoverability of Validator infrastructure," including "hardened configurations, … backup and recovery, … supply-chain security, or security controls for hosted Validator services." Against each of the RFP's terms:
+
+* *Hardened configuration and security controls for hosted Validators.* The participant runs inside a TEE; the hosting operator cannot read its state, alter it, or sign in its name, and has neither direct full access to the Ledger API nor unrestricted access to debug logs (§6).
+
+* *Backup and recovery.* State persists outside the enclave as encrypted snapshots under a sealed persistence key that only an attested build can unseal; a replacement enclave restores from the latest snapshot with no data loss (M2).
```
simran-da, line 142: what exactly is meant by a replacement enclave?

rfps/security-assurance-incident-readiness/2026-08-FPI-TEE-covalidator.md
```diff
@@ -0,0 +1,228 @@
...
+
+The work operates within CIP-0100 governance.
+
+**4\. Backward Compatibility**
+
+Purely additive. A TEE co-validator joins a party's hosting set like any participant. No protocol changes; no changes required from other nodes; removal is an ordinary topology change. A natural follow-on, outside this proposal's scope, is protocol-level recognition of TEE attestations, so that the network itself, rather than the key provisioner, blesses valid enclave builds. We would be glad to draft that as a separate proposal once this model has proven itself.
+
+**5\. Software upgrades**
+
+Operators are able to upgrade Splice to newest versions without a complicated process involving direct involvement of a Party. We assume trust to Splice binaries signed by Digital Asset. Based on this assumption, the harness inside an enclave will accept an upgrade command from an Operator and switch running Splice binaries internally. This is done without a Party directly authorising an update.
```
simran-da, line 160: do I understand correctly that updates will roughly proceed as follows: the node operator downloads the new binary, builds the new enclave image (.eif) and computes its new PCR0 hash -> operator updates AWS KMS policy and attestation framework to legitimize this new hash for use -> operator boots up the new enclave alongside the old one -> once the new enclave successfully takes over the workload, the operator safely destroys the old enclave and updates the AWS KMS policy again to remove the old PCR0 hash? how would we integrate a verification of DA's software signature on the new binary into this setup?

rfps/security-assurance-incident-readiness/2026-08-FPI-TEE-covalidator.md
```diff
@@ -0,0 +1,228 @@
...
+    classDef encl fill:#FCFDFC,stroke:#35566B,color:#1A1D1F
+    classDef note fill:none,stroke:none,color:#5D666E,font-style:italic
+    class NOTE note
+    class SP,PG,NET,ST solid
+    class IK,PK,AK,AH key
+    class EC2 outer
+    class ENC encl
+```
+
+**Figure 2** — Three keys, three jobs. The identity key never leaves the enclave; the persistence key returns only to a build with matching measurements; the audit key is an optional, operator-configured interface whose private half sits wherever the operator chooses. The operator of the EC2 parent instance sees ciphertext and vsock traffic only.
```
simran-da, line 101: have you done an approximate estimation of the added performance load of proxying all network and DB calls through vsock? (There are also the added DB encryptions and decryptions, but these are likely negligible for our case). Probably this number is not significant compared to Canton's processing time, but it would be good to have a rough idea nevertheless.

rfps/security-assurance-incident-readiness/2026-08-FPI-TEE-covalidator.md
```diff
@@ -0,0 +1,228 @@
...
+
+Out of scope: compliance and disclosure logic, analytics integrations, and key-custody services. These are downstream consumers of the open interfaces defined here, not part of this proposal.
+
+**2\. Implementation Mechanics**
+
+**Packaging.** We run a Splice participant and its PostgreSQL store as a single multi-process image inside a TEE Enclave. It is wrapped by a harness running inside an enclave, developed by FPI. It manages enclave access and acts as a proxy for certain operations. The deliverable is that image, host-side setup scripts (docker-compose and Terraform), and documentation; standard host components such as the vsock proxy are documented rather than shipped. A working prototype exists: the enclave-packaged node has run against a simulated Canton network, and we separately operate standard, non-enclave validator and TestNet participant nodes. That operating experience feeds directly into the tooling and runbooks this proposal delivers. Milestone 1 merges the two tracks: the enclave-packaged node synchronizing DevNet.
```
simran-da, line 53: Assuming we agree to proceed with placing the DB encrypted in the parent host, how would Canton's JDBC database calls and SQL queries need to be changed? My gut feeling is that this may get quite complicated, even if we encrypt only the transaction payloads. Is there something else we could do to keep things seamless and still avoid any decryption on the parent host? A very quick search yielded something like this https://docs.anjuna.io/latest/nitro/v1.58/getting_started/persistent_storage.html - is there a chance we could implement our own minimal version of this?

rfps/security-assurance-incident-readiness/2026-08-FPI-TEE-covalidator.md
```diff
@@ -0,0 +1,138 @@
...
```
alexeykoren: First step would indeed be a in-enclave database that have size limitations on top of scaling, etc.

Second iteration we are going to implement database, that runs inside enclave, but stores its data files outside in encrypted file system. This solves size problem and also helps with the problem of enclave restart. We are going to evaluate performance/reliability/scale to understand if its crucial or not to do application level encryption

rfps/security-assurance-incident-readiness/2026-08-FPI-TEE-covalidator.md
```diff
@@ -0,0 +1,138 @@
...
+
+**5\. Software upgrades**
+
+Operators are able to upgrade Splice to newest versions without a complicated process involving direct involvement of a Party. We assume trust to Splice binaries signed by Digital Asset. Based on this assumption, the harness inside an enclave will accept an upgrade command from an Operator and switch running Splice binaries internally. This is done without a Party directly authorising an update. 
+
+**6\. Limited access to the Canton node**
+
+Operators don’t have direct full access to Ledger API to protect Party’s privacy. Access to debug logs is also limited. For cases when selective access is required, e.g. for mapping a Party to a participant \- there is a command in harness that will deliver required call data and signatures to the node. This may include, for example, enabling temporary access to debug logging for operators.
```
alexeykoren: It is going to be a wrapper/proxy, techically. Although we are going to keep formats as close to original API as possible

rfps/security-assurance-incident-readiness/2026-08-FPI-TEE-covalidator.md
```diff
@@ -0,0 +1,138 @@
...
+
+**3\. Architectural Alignment**
+
+Functionally, a TEE co-validator behaves exactly like any confirming participant in the existing co-validation model. It hosts parties, validates its view, and counts toward confirmation thresholds. Protocol, standards, and topologies are unchanged. The difference is the trust boundary: the node's operator sits outside it, so co-validation no longer requires showing another organization the party's transactions. Canton gains cooperative validation that preserves financial privacy. The work fits the fund's stated scope under CIP-0082 (security, reference implementations, critical infrastructure) and operates within CIP-0100 governance.
+
+**4\. Backward Compatibility**
+
+Purely additive. A TEE co-validator joins a party's hosting set like any participant. No protocol changes; no changes required from other nodes; removal is an ordinary topology change. A natural follow-on, outside this proposal's scope, is protocol-level recognition of TEE attestations, so that the network itself, rather than the key provisioner, blesses valid enclave builds. We would be glad to draft that as a separate proposal once this model has proven itself.
```
alexeykoren: Its going to be the same harness (wrapper/proxy) that is exposed via vsock. It will handle admin access with admin's identity being part of enclave hash for meximum security

proposals/2026-08-FPI-TEE-covalidator.md
alexeykoren: Enclaves can be several hundred of GBs big, which is big enough for many usecases except most traffic-heavy. We are running performance tests to calculate a practical limit and state it clearly in documentation

rfps/security-assurance-incident-readiness/2026-08-FPI-TEE-covalidator.md
```diff
@@ -0,0 +1,204 @@
...
+            NOTE ~~~ SP
+            SP ~~~ IK
+            PG ~~~ PK
+        end
+    end
+    NET["<b>CANTON NETWORK</b><br/>counts the confirmations"]
+    ST["<b>EXTERNAL STORAGE</b><br/>ciphertext only<br/>snapshots · exports"]
+    AH["<b>SK_audit HOLDER</b><br/>self-custody (default) · HSM · service provider"]
+    IK -- confirmations --> NET
+    PK -- snapshots --> ST
```
alexeykoren: Yes, postgres is supposed to backup its state automatically and periodically as if it was a normal non-cloud db

rfps/security-assurance-incident-readiness/2026-08-FPI-TEE-covalidator.md
```diff
@@ -0,0 +1,204 @@
...
+    ST -. restore on restart .-> PG
+    AK -- exports --> ST
```
alexeykoren: Its actually the same in this context, I'll fix it to use the same terminology

rfps/security-assurance-incident-readiness/2026-08-FPI-TEE-covalidator.md
```diff
@@ -0,0 +1,204 @@
...
+
+**Attestation.** We build reproducibly, publish PCR0/1/2 per release, and provide a verification guide so any third party can rebuild the image from source and check the measurements against a running enclave’s attestation document. Trust in the node reduces to trust in published, reviewable code.  We use Splice binaries signed by Digital Asset as a core building block for reproducible builds. 
+
+**Keys.** Three keys, with three distinct jobs:
+
+* *Canton identity key.* The co-validator's signing key, the identity whose confirmations the network counts. It is provisioned into the enclave and never leaves it. This is what makes the co-validation itself trustworthy: if this key were available outside the TEE, its holder could sign confirmations from anywhere.
```
alexeykoren: Indeed, the first one

As for management non-root keys our plan is to expose enclave API for doing that with auth only required from node operator side for simpler operations

rfps/security-assurance-incident-readiness/2026-08-FPI-TEE-covalidator.md
alexeykoren: Exactly, there will be an enclave API that returns fresh attestation based on request (challenge)

rfps/security-assurance-incident-readiness/2026-08-FPI-TEE-covalidator.md
alexeykoren: Node operator never holds private key in plaintext normally. Here we have an external party (auditor), that provides their public key to be bundled into enclave's PCR0. The way how auditor proves that data will not be misused is no concern of Canton enclave and depends on auditor, but we envision auditor analytics to run inside of their own enclave and attestaion of that enclaves proves that analytics can only be done within agreed bounds with proper logging of access

rfps/security-assurance-incident-readiness/2026-08-FPI-TEE-covalidator.md
alexeykoren: A new instance of the same codebase that we spin out instead of the one that went down

rfps/security-assurance-incident-readiness/2026-08-FPI-TEE-covalidator.md
alexeykoren: This is one of the ways, that definitely will be available. In this case correctness of the binary will be verified during EIF building step

Alternatively we are also looking at the option of uploading a new binary inside running enclave. In this case harness will restart node wit ha new binary and a DA signature over its hash will be part of the API, while DA's public key will be a part of original config (and PCR0)

rfps/security-assurance-incident-readiness/2026-08-FPI-TEE-covalidator.md
alexeykoren: We are working on it right now, it depends a little bit on the way how db is set up, but in general vsock round is indeed negligible comparing to Canton processing times, projected to be under a millisecond. Encrypted filesystem will add to that, but we still expect that to be very low overhead

rfps/security-assurance-incident-readiness/2026-08-FPI-TEE-covalidator.md
alexeykoren: Application level encryption is indeed a thing that can become quite big in terms of chnages

But in case of encrypted filesystem JDBC will connect to a vanilla PG, without knowing its internal which requires no change in Splice at all

PR #184, opened 2026-04-10
Proposal: Concordia

luis-marado-unlockit:
## Development Fund Proposal Submission

**Proposal file:**  
`/proposals/concordia.md`

---

## Summary
This proposal submits Concordia, an open-source reference implementation for privacy-preserving multi-party allocation and decision workflows on Canton. It delivers a reusable core for submission, aggregation, settlement, and expiry handling, plus governance and auction modules with reference applications and documentation for reuse and extension.

---

## Checklist
- [x] Proposal file added under `/proposals/`
- [x] Milestones and funding amounts defined
- [x] Acceptance criteria included
- [x] Alignment with Canton priorities described

---

## Notes for Reviewers
This proposal is championed by IntellectEU and Jonathan Mayeur.

This proposal is intended to contribute reusable ecosystem infrastructure.

The first release focuses on a bounded but extensible CAP core and two proving domains: governance and auctions. The core is designed to sit above existing Canton asset and settlement infrastructure so teams can reuse allocation and decision mechanics without replacing the token and transfer layers they already depend on.

The reference applications and documentation are intended to make the implementation practical for external Canton teams to evaluate, adopt, and extend.

PR #741, opened 2026-08-28
Proposal: Anode to design permissioned Governance system

raphbaph:

**Proposal file:**
[2026-08-anode-governance-permission-design.md](https://github.com/raphbaph/canton-dev-fund/blob/codex/anode-governance-permission-rfp/proposals/2026-08-anode-governance-permission-design.md)

### Applicant

**Organization:** Anode.GG

**Author / Primary Contact:** Raphael Spannocchi, Anode.GG

**Champion:** Heslin Kim, Zenith

---

### Proposal Classification

**Proposal Type:**
- [x] RFP-aligned proposal
- [ ] Individual initiative

**RFP / Roadmap Area:** RFP 9 — Governance automation

**Label:** onchain-governance

---

### Funding & Timeline

**Total Funding Request:** 860,000 CC

**Project Duration:** 6–9 months

**Maximum Amount:** N/A

**Maximum Duration:** N/A

---

### Summary

Anode.GG proposes an open-source reference implementation for permissioned, typed Canton governance proposals. It provides a mandate registry, submission gate, and lifecycle/audit workflow so a proposal is attributable to valid, time-bound authority before it reaches existing governance channels—reducing manual overhead without changing voting rights, voting weight, quorum, or the one-vote-per-node model.

---

### Submission Checklist

- [x] Full proposal file is included in this PR
- [x] Organization and primary contact identified
- [x] Champion identified
- [x] RFP / roadmap alignment identified
- [x] Total funding request provided
- [x] Project duration provided
- [x] Proposal is within any RFP maximum amount
- [x] Proposal is within any RFP maximum duration
- [x] Milestones and milestone funding are defined in the proposal
- [x] Acceptance criteria are based on ecosystem value
- [x] Architectural alignment is addressed

### DIFF
--- a/proposals/2026-08-anode-governance-permission-design.md
+++ b/proposals/2026-08-anode-governance-permission-design.md
@@ -1,0 +1,294 @@
+# Development Fund Proposal
+
+| Field | Value |
+| :---- | :---- |
+| **Organization** | Anode Governance |
+| **Author / Primary Contact** | Raphaël Baph, Anode Governance |
+| **Status** | Submitted |
+| **Created** | 2026-08-26 |
+| **Proposal Type** | RFP-aligned |
+| **RFP / Roadmap Area** | RFP 9 — Governance automation |
+| **Champion** | Needs Champion |
+| **Total Funding Request** | 430,000 CC and 47,000 USDC |
+| **Project Duration** | 6 months |
+| **Label** | onchain-governance |
+
+---
+
+## Abstract
+
+Anode Governance proposes an open-source reference implementation for **permissioned, typed governance proposals** on Canton. The project introduces an off-chain mandate registry and submission gate that permits a person or body to submit only the decision classes that Super Validators have explicitly authorized, within recorded limits and expiry dates. It reduces manual proposal triage, makes authority traceable, and gives Super Validators a more predictable review workflow without changing Canton’s current voting weights, quorum, or voting mechanism.
+
+The first release implements the source proposal’s **Option B**: proposals remain subject to the existing voting process, but they are correctly typed, complete, attributable to a valid mandate, and bundled into predictable review windows. The architecture will keep a later consent-agenda path and professional-proxy workflow possible, but neither is in scope for this grant.
+
+---
+
+## Specification
+
+### 1. Objective
+
+Deliver a reusable governance-permissioning reference implementation that makes proposal authority explicit before a proposal reaches Canton’s existing governance process.
+
+Today, governance involves numerous committees, working groups, editors, Super Validator operators, and Special Interest Groups (SIGs). The authority boundaries between them are often implicit. At the same time, a broad range of decisions arrive through the same high-attention Super Validator voting process. Rate limits can reduce submission volume, but cannot establish whether a proposer is authorized to make a particular proposal.
+
+The project addresses that gap with three connected capabilities:
+
+1. A machine-readable **mandate registry** that records who is authorized to initiate which decision classes, subject to stated limits, escalation rules, provenance, and expiry.
+2. A **typed submission gate** that validates a proposal against its mandate, rejects out-of-scope submissions before publication, and records the authority chain for accepted proposals.
+3. A transparent **workflow and audit layer** that tracks notice periods, review windows, renewals, notifications, and the complete provenance trail from a decision back to the mandate and authorizing vote.
+
+The objective is deliberately limited to proposal creation and governance workflow coordination. It does not replace the existing voting mechanism or determine who receives a vote.
+
+### 2. Problem and Ecosystem Value
+
+For institutional participants, governance needs a documented chain of responsibility. A committee, working group, or delegated role should not derive its authority merely from custom or an informal process. It should hold a visible mandate that states what it may initiate, under which constraints, and until when.
+
+This is consistent with institutional governance practice:
+
+| Institutional practice | Canton capability delivered by this proposal |
+| :---- | :---- |
+| Board delegation of authority | Time-bound mandate with defined decision classes and limits |
+| Reserved matters | Decision classes that always route to a full Super Validator vote |
+| Consent agenda | A future-compatible workflow for routine items; not enabled in the initial release |
+| Annual general meeting / reauthorization | Expiring mandates that can be renewed in a bundled review cycle |
+| Delegation register | Public, queryable mandate registry and provenance trail |
+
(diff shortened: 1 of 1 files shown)


waynecollier-da:
SV governance has a pretty predictable flow right now. There are well-defined agreements about governance procedures and operational expectations, and there are no examples since MainNet launch of conflicts or disputes about the process. The Foundation coordinates all requests for votes; the SV Voting Group reviews and acts on all proposals, and onchain votes begin only after a proposal has been approved offchain by the SV Voting Group. 

I don't see a need for this proposal. 

Further, the current governance priorities outlined by the Super Validators are: 

1. Implement Rewards locking (Obsidian is working on a grant here). 
2. Get SV weights onchain (IntellectEU has submitted a proposal for this). 
3. Move the SV Governance app to a dApp architecture with external signing (Avro Digital's work). 
4. Introduce weighted voting, and expand the voting pool to all SV rights owners (This CIP is still pending consensus in cip-discuss). Implementation of step 4 will need to wait for steps 2 and 3 to complete. 

Your proposal might fit after step 4 is complete, because the size of the voting pool will expand significantly, the role of the SV Voting group will change, and new governance procedures will be required. 

Best regards, 

Wayne Collier

heslin-zenith:
Hi Wayne,

Thank you for taking the time to write this out, and for laying out the current priorities so clearly. I agree with you on the immediate roadmap: rewards locking, onchain SV weights, the move to a dApp with external signing, and then a careful expansion of the voting pool. Those items should stay first.

I also take your point that SV governance has been orderly since MainNet. That is real, and it is to everyone’s credit. I championed this proposal, because my concern is not that the present process has failed but that the process we have today is still largely social, and social processes tend to hold until the set of participants, interests, and proposal volume changes. The work now being sequenced will change all three.

That is why I think this proposal is worth considering in parallel, rather than only after step 4.

What the proposal actually does is narrow. It’s not seeking to change voting weights, quorum, or who casts a vote. It’s not replacing the SV Voting Group, the Foundation’s coordination role, or Avro’s dApp work. We have a missing element of checks and balances in the current model. This design, as explained to me, settles who is authorized to initiate which class of proposal, within what bounds, and until when in a standardized and publicly outlined approach. This is how we can settle routine operational items and easing the workload of the few existing members by offloading it to the whole. 

The value of this is preventative. Disputes we’ve needed to engage in as Zenith with current governance were far before the onchain vote. They began earlier, in non-transparent, “async meetings” when people disagree about whether a proposal was in order, who was entitled to bring it, or whether an informal understanding still applied. While the group is small and well aligned, those questions can be resolved by direct conversation. As the voting pool grows, and as the role of the SV Voting Group changes, those same questions become harder and more political if they are still settled by existing custom. Putting authority in a public, time-bounded register is a way to keep future disagreement settled by neutrality.

I also want to be candid about how we talk about weighted voting. I am not opposed to weight, locking, or long-term commitment. Those tools can be useful. But weight derived from accumulated rewards, locked stock, or large treasury positions will, by design, give more influence to parties that already hold more. That may be the trade “we” choose. We should just be clear-eyed that it concentrates agenda-setting power unless the front end of governance with who may propose what is explicit and reviewable. Expanding the pool without first clarifying proposal authority would make that concentration easier to miss, not easier to manage.

A related structural point, offered in the same spirit: it remains difficult for later SVs to enter the infrastructure-operator set on comparable terms. Early operators also earned under a much higher reward regime than anyone joining now. Combined with locking and weight, that creates a durable advantage for incumbents even where no one intends one. I raise this point because governance design should not rely on goodwill alone once incentives and history have already produced unequal starting positions. Clearer mandates and a documented path for operational participation would reduce the risk that later participants experience the system as closed.

I do not see this proposal as a substitute for the four priorities you listed, and I would not ask that it jump the queue ahead of them. I see it as complementary plumbing. If step 4 is going to enlarge the electorate and change the SV Voting Group’s role, it will be easier to do that well if the language of authority around decision class, bounds, expiry, and escalation already exists. Building that language after the pool expands is possible, but I championed this because I believe building it before is preventative.

I needed to justify my decision to supporting this here as I’ve had a different experience from your own vantage point. 

Best regards,  
Heslin Kim
Zenith

waynecollier-da:
Hi, Heslin

Based on the proposal content, and your response, I'd recommend that this proposal be preceded by a CIP outlining the proposed changes to early-stage governance processes directly, and calling for agreement. It would be fine to point to this proposal as an argument that there's a technical plan in place for implementing this type of early-stage governance if it there were agreement to adopt it. But this sounds more like a matter of conceptual alignment than something that's ready for an implementation.

raphbaph:
Hi Wayne,

Thanks for being open to this. We've started drafting a CIP and will publish it shortly.
One option we did not include in the proposal draft here was that of delegated voting. The call for delegates has come up in cip-discuss.

In our opinion, Canton governance should be shaped by how corporate governance runs now. This is what institutional SVs already are familiar with, which reduces any compliance work necessary on their end wrt voting.

This is also part and parcel of the current proposal.

Delegation could be modelled after proxy voting for joint stock corporations, where proxies have mandates allowing them certain actions for certain decision classes, but also accept potential recourse for decisions.

we have a draft for this, and will likely submit this separately to move the discussion ahead.

luis-marado-unlockit:
Hi @raphbaph, thanks for sharing this on #sig-onchain-governance. From @heslin-zenith’s explanation, the governance clearinghouse sounds closely related to use cases that our clients are already implementing on top of Concordia’s primitives.

In these implementations, governance rules are themselves outcomes of governed processes. Once approved, those rules determine who may initiate particular classes of proposals, the applicable bounds and expiry, and the decision thresholds or authorization requirements for each category.

Members can subsequently propose amendments to those rules through Concordia. Different proposal categories can therefore follow different thresholds based on previously approved outcomes, allowing the governance model to evolve through explicit, auditable decisions rather than informal coordination or fixed application logic.

Concordia already provides reusable primitives for proposal submission, deterministic resolution, authorization, expiry, auditable outcomes, and execution. The clearinghouse could therefore be implemented as a governance model on top of Concordia rather than as a separate workflow and execution framework.

Have you considered reusing Concordia for this? In particular, it would be useful to understand how mandates and proposal categories would be created, amended, revoked, and enforced. This would help clarify how the clearinghouse could extend the existing primitives while avoiding duplication of the underlying governance foundation.

Proposal: https://github.com/canton-foundation/canton-dev-fund/pull/184  
Active development: https://github.com/unlockitio/concordia

There may also be an interoperability point with the #sig-identity-and-metadata group’s work on service discovery, particularly if governance decisions and the rules established by their outcomes need to be published, discovered, and verified through a governance registry.

PR #53, opened 2026-03-06
Development Fund Proposal: ISS-Based BFT

waynecollier-da:
## Development Fund Proposal Submission

**Proposal file:**  
Link to the proposal added in this PR (e.g., /proposals/your-proposal-name.md)

---

## Summary
Briefly describe the proposal and the value it delivers to the Canton ecosystem (2–3 sentences).

---

## Checklist
- [ ] Proposal file added under `/proposals/`
- [ ] Milestones and funding amounts defined
- [ ] Acceptance criteria included
- [ ] Alignment with Canton priorities described

---

## Notes for Reviewers
(Add anything the Tech & Ops Committee should pay attention to.)

PR #750, opened 2026-08-31
[Amendment] ISS-Based BFT #53 M4: Automatic Pruning & Operational Maturity

guillaumedejenlis-da:
Digital Asset requests a scope change to continue work on Grant #53. At the time of the grant writing, pruning was managed manually by each SV. At the current time, the 30 days pruning window is sufficient for SVs and there is no need to make additional improvements to data pruning. 

Therefore, we have taken the decision to descope this feature and allow for additional performance improvements allowing stable and fast deployment of the BFT including: 
- Performance and stability improvements, especially for networks with heterogenous load and latency
- Nodes fetch batch data proactively and simultaneously from multiple peers to reduce delivery and ordering latency
- Improved peer connection management to avoid requesting data from nodes that recently timed out or became disconnected
- Nodes flush remaining segment blocks if a sufficient number of peers have already finished their own respective segments
- Nodes use a dynamic view change timeout to increase the allotted time for new leaders (in views > 0) to successfully order blocks
- Reduced memory footprint of expensive view changes: unnecessary messages are removed from memory, and across restarts, nodes load only the precise messages needed for rehydration
- Exponential backoff variant of leader blacklisting to more quickly restore long periods of fast ordering when a failed leader experiences a persistent problem
- Holistic database partitioning scheme to improve the performance and efficiency of periodic pruning. And in general, improved the efficiency of many database queries to use indexes and avoid sequential scans.
- Enhanced the dynamic sequencing parameters to enable network-wide consistency of additional important properties, including view change settings, leader blacklisting heuristics, and maximum batch sizes for blocks

### DIFF
--- a/proposals/2026-02-DA-ISS-based-BFT.md
+++ b/proposals/2026-02-DA-ISS-based-BFT.md
@@ -153,15 +153,23 @@ Deliverables:
 - Throughput Demonstration  
 - Governance Demonstration
 
-## Milestone 4: Automatic Pruning & Operational Maturity
+## Milestone 4: Performance & Stability
 
 | Estimated Delivery | Month 5 (August 2026 if approved in March) |
 | :---- | :---- |
-| **Focus** | Automatic distributed pruning, operational tooling, documentation |
+| **Focus** | Performance and stability for heterogeneous networks, operational tooling, documentation |
 
 Deliverables:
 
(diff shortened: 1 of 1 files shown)


shaul-da:
Update from the meeting: overall looks good but people haven't reviewed sufficiently so deferred to next meeting

heslin-zenith:
Seems reasonable to me with the adjusted work scope. Zenith in favor of the alteration.

hythloda (approved):
Approved via tech-ops voting meeting on Sept 16th, 2026

PR #759, opened 2026-09-03
Proposal/canton reserve attestation

amaramci:
Development Fund Proposal Submission

Proposal file:
/proposals/cantonreserveattestation.md

Summary

Canton Reserve Attestation is an open-source Daml reference implementation that lets a custodian prove solvency continuously and cryptographically, rather than through periodic snapshots or trust-based attestations. For assets and client claims that are native Canton contracts, a withdrawal that would breach the solvency invariant is structurally rejected by the ledger itself. Several custodians already building on Canton (BitGo, Copper, Dfns) are the concrete adoption target for this reference implementation.

### DIFF
--- a/proposals/cantonreserveattestation.md
+++ b/proposals/cantonreserveattestation.md
@@ -1,0 +1,107 @@
+## Canton Reserve Attestation: Structural Proof-of-Reserves for Custodians
+
+**Author:** Amar Mujezinovic (amarmujezinovic1@gmail.com)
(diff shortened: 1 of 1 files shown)


Jatinp26:
Review:
- Some technical rework needed in the ClaimAggregate snapshot section.
- Adoption and need for this in current ecosystem is not clear at the moment.
- Need some more evidence to support the need in the ecosystem.

